PR #24, opened 2018-05-29
Adds enum capability -> fixes #21

### LINKED ISSUE #21: Add simple enum validator (that can be any of type)
Hi! Thanks for your great tool) 
I think "enum" validator will be very userful. At this time this is available for string validator. But, for example, I often use numeric enums in REST API as parameters.
I know that it can be made by using multiple validators. But that will be great to write like:
```
let schema = {
    status: { type: "enum", values: [null, 1, 2, "done"] }
}
v.validate({ status: "done" }, schema)
```

mauricedoepke:
I accidently mixed this with fix  for 22 before.


icebob:
Released, but I've added the missing `actual` to error object and missing message text :)

PR #34, opened 2018-10-26
Support recursive schemas

### LINKED ISSUE #30: Recursive schema
Hi, trying to validate a recursive schema for f.ex a category tree, but errors during v.compile with "Maximum call stack size exceeded". 

```js
describe("Test recursive schema", () => {
	const v = new Validator();

	let schema = {};
	Object.assign(schema, {
		name: { type: "string" },
		subcategories: {
			type: "array",
			optional: true,
			items: { type: "object", props: schema}
		}
	});

	let check = v.compile(schema);

	it("should compile", () => {
		let category = {
			name: "top",
			subcategories: [
				{
					name: "sub1"
				},
				{
					name: "sub2"
				}
			]
		};

		let res = check(category);

		expect(res).toBe(true);
	});
});
```

andersnm:
Fixes #30

Currently the cache holds references to all schema rules for the lifetime of the validator. Should probably clear the cache after each compile to prevent leaks in long-running programs, but haven't made up my mind yet.

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -39,6 +39,7 @@ function Validator(opts) {
 
 	// Load rules
 	this.rules = loadRules();
+	this.cache = new Map();
 }
 
 /**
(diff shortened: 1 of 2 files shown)


icebob:
Is it cut down the performance?

andersnm:
It does not affect the performance of the compiled validators. It is just doing some extra checks during the compilation step.

andersnm:
I've reorganized the PR so it handles both schemas and input data with cycles. And added a test where recursive schema produces error.

Turns out this does not affect the performance of existing non-cyclic schemas (if any, I'm getting slight improvements!). Although object types with cycles validate somewhat slower than non-cyclic ones.

icebob:
Thanks!

PR #36, opened 2018-11-02
Update readme

zmarouf:


zmarouf:
I did a quick pass. There might still be things to change. I'll check the code to replace _larger_ with _greater_

zmarouf:
Never used Codacy but is this normal?
image

PR #37, opened 2018-11-02
Handle irregular object property names

### LINKED ISSUE #35: v0.6.11 broke some specific schemas
I use UUID literals nested keys in validation schema its work on 0.6.10 and was broken in 0.6.11.
Simple example code:
```js
const schema = {
  minPrice: {
    type: 'object',
    props: {
      value: { type: 'number', positive: true, integer: true },
      enabled: 'boolean',
      regions: {
        type: 'object',
        optional: true,
        props: { '12345678-ef85-4c62-b8f3-e00fbbc155d4': { type: 'number', positive: true, integer: true, optional: true } },
      },
    },
  },
};

/**
 * Next line throws
 * SyntaxError: Invalid or unexpected token
 * at new Function (<anonymous>)
 * at Validator.compileSchemaObject (node_modules\fastest-validator\lib\validator.js:114:33)
 * at Validator.compileSchemaRule (node_modules\fastest-validator\lib\validator.js:154:24)
 * at Validator.compileSchemaType (ec-order\node_modules\fastest-validator\dist\index.js:472:15)
 *
 * Generated code looks like
 * propertyPath = (path !== undefined ? path + ".12345678-ef85-4c62-b8f3-e00fbbc155d4" : "12345678-ef85-4c62-b8f3-e00fbbc155d4");
 * res = this.checkSchemaRule(value.12345678-ef85-4c62-b8f3-e00fbbc155d4, compiledObject[0].compiledType, propertyPath, value);
 * if (res !== true) {             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 *  this.handleResult(errors, propertyPath, res);
 * }
 */
const result = v.validate(data, schema);
```

At the moment, I don't have enough time to make a pull request.

andersnm:
Fixes #35 

Two notes:
- No particular handling of reserved keywords as property names. They are supported since ES5, and the codegen already uses ES6/ES2015 keywords like `const`
- Does not optimize all valid unquoted Unicode property names allowed by the spec. This requires a Unicode library or including some large-ish tables

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -22,6 +22,34 @@ function loadRules() {
 	};
 }
 
+// Quick regex to match most common unquoted JavaScript property names. Note the spec allows Unicode letters.
+// Unmatched property names will be quoted and validate slighly slower. https://www.ecma-international.org/ecma-262/5.1/#sec-7.6
+const identifierRegex = /^[_$a-zA-Z][_$a-zA-Z0-9]*$/;
+
+// Regex to escape quoted property names for eval/new Function
+const escapeEvalRegex = /["'\\\n\r\u2028\u2029]/g;
+
+function escapeEvalString(str) {
+	// Based on https://github.com/joliss/js-string-escape
+	return str.replace(escapeEvalRegex, function (character) {
+		switch (character) {
+		  case '"':
+		  case "'":
+		  case '\\':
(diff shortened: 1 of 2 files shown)


icebob:
So I should make a new release?

Issue #18, opened 2018-04-04
Add sanitizers function


icebob:
## Sanitizers

- [ ] **convert strings to numbers**
  ```js
  age: { type: "number", convert: true }
  ```
- [ ] **convert strings & numbers to booleans**
  ```js
  age: { type: "boolean", convert: true }
  ```
- [ ] **convert strings to dates**
  ```js
  age: { type: "date", convert: true }
  ```
- [ ] **trim strings**
  ```js
  name: { type: "string", trim: true }
  ```
- [ ] **left-trim strings**
  ```js
  name: { type: "string", ltrim: true }
  ```
- [ ] **right-trim strings**
  ```js
  name: { type: "string", rtrim: true }
  ```
- [ ] **lowercase strings**
  ```js
  name: { type: "string", lowercase: true }
  ```
- [ ] **uppercase strings**
  ```js
  name: { type: "string", uppercase: true }
  ```
- [ ] **normalize emails** (lowercase, ...)
  ```js
  name: { type: "email", normalize: true }
  ```
- [ ] **default value** (if `undefined`)
  ```js
  role: { type: "string", optional: true, default: "user" }
  ```

icebob:
https://github.com/icebob/fastest-validator/pull/94

Issue #15, opened 2018-02-14
convert true to boolean don't actual converts the value.


prieston:
Convert boolean values as I see from the boolean.js lib it only returns true and not converting the value

icebob:
Yes, currently it is just a validation. Don't modify the values. Only convert values for checking.

Prieston:
Ok so will it be supported in future releases?

icebob:
I think yes. it will be implemented.

andersnm:
What could this look like, API-wise? And should it mutate or clone the input?

My WIP sanitize branch adds a new parameter to the validator functions, where a sanitized clone of the input is returned. F.ex:

```js
const v = new Validator();
const converted = {};
const res = v.validate(data, schema, converted);
// the sanitized data is now in converted.value
```

This leaves the API backwards compatible and mostly unchanged. A maybe nicer, but breaking change could use the return value:

```js
const v = new Validator();
const errors = [];
const converted = v.validate(data, schema, errors);
// the sanitized data is now in converted
// if converted is null, validation failed and the errors array contains the error messages.
```

Both scenarios handle when the input is a primitive type like string or number.

PR #44, opened 2018-11-09
Implement to return sanitized object

andersnm:
This proposes to add an optional `converted` parameter to the compiled checker function whose `value` property will be assigned the converted/sanitized value. It provides the framework for #15, upon which #18 could be implemented.

The performance is down somewhat because of the additional checks (~4.1mil->3.9 mil validations/sec on my box). This could be mitigated by having different code if the converted parameter is provided or not, but I have not done this (yet?).

I assume Codacy will complain again, but I dont know how to run the same checks locally. :(

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
 			"request": "launch",
 			"name": "Launch benchmark",
 			"program": "${workspaceRoot}\\benchmark\\index.js",
-			"args": [ "simple" ]
+			"args": [ "sanitize" ]
 		},
 		{
 			"type": "node",
--- a/benchmark/suites/sanitize.js
+++ b/benchmark/suites/sanitize.js
@@ -1,0 +1,74 @@
+"use strict";
+
+const Benchmarkify = require("benchmarkify");
+const benchmark = new Benchmarkify("Fastest validator benchmark").printHeader();
+
+let bench = benchmark.createSuite("Sanitize simple object");
+
+const Validator = require("../../index");
+const v = new Validator();
+
+const obj = {
+	name: "John Doe",
+	email: "john.doe@company.space",
+	firstName: "John",
+	phone: "123-4567",
+	age: "33",
+	status: 1
+};
+
+const schema = {
+	name: {
+		type: "string",
+		min: 4,
+		max: 25
+	},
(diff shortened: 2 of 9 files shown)


icebob:
Good job! Codacy said there are some indent issues only. No any serious problem.
How much work to create a different code for converted code and without it?

icebob:
@andersnm ?

andersnm:
@icebob I did commit some additional optimizations to restore the performance to pre-sanitizer for both objects and arrays :-). I'm not super happy about the API, but I think its the best we can do without breakage, and hopefully no bugs/undesired side effects.

icebob:
Thanks for your work. I've started to check the repo and make benchmark tests.
Yes, the "converted" solution a little bit strange, but we can change it in the future at v1.x.x

icebob:
Hi @andersnm,
I've added a [benchmark test for sanitization](https://github.com/andersnm/fastest-validator/blob/sanitize/benchmark/suites/sanitize.js), but it throw error. Do you know what is wrong with benchmark?

```
Suite: Sanitize simple object
× [ERR] validate & sanitize
ReferenceError: propertyValue is not defined
    at Validator.eval (eval at Validator.compileSchemaObject (D:\Work\tmp\fastest-validator\lib\validator.js:139:42), <anonymous>:20:18)
    at Validator.checkSchemaObjectInner (D:\Work\tmp\fastest-validator\lib\validator.js:261:47)
    at Validator.checkSchemaObject (D:\Work\tmp\fastest-validator\lib\validator.js:256:15)
    at D:\Work\tmp\fastest-validator\lib\validator.js:108:15
    at TestCase.bench.add [as fn] (D:\Work\tmp\fastest-validator\benchmark\suites\sanitize.js:42:12)
```

andersnm:
It's a bug, and found another bug as I wrote a test for it. :( Sorry, will have to get back to you later.

andersnm:
I've fixed the bugs I found and added tests locally, but lost steam upon realizing there should be far more tests for the sanitizing feature. Can try and commit what I have later this week

andersnm:
It has been rebased and force-pushed

icebob:
https://github.com/icebob/fastest-validator/pull/94

PR #47, opened 2018-11-12
Strict object validation

fabioanderegg:
I propose the addition of strict object validation:
Validate objects to only allow fields that are referenced in the props dictionary of an object.
```js
const schema = {
    obj: {
        type: "object",
        props: {
            a: "string"
        },
        strict: true
    }
};

console.log(v.validate({ obj: { a: "string" } }, schema));
// Returns: true

console.log(v.validate({ obj: { a: "string", b: "string" } }, schema));
/* Returns an array with errors:
[ { type: 'objectStrict',
    expected: 'b',
    actual: undefined,
    field: 'obj',
    message: 'The object \'obj\' contains invalid keys: b!' } ]
*/
```

Perhaps someone has a better name than "strict object validation" for this.

The performance impact with `npm run bench` is not measurable, but adding a separate "strictObject" validator is certainly possible if this is a problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
 * 13 built-in validators
 * custom validators
 * nested objects & array handling
+* strict object validation
 * multiple validators
 * customizable error messages
 * programmable error object
@@ -48,7 +49,7 @@ You can install it via [NPM](http://npmjs.org/).
 ```
 $ npm install fastest-validator --save
 ```
-or 
+or
 ```
 $ yarn add fastest-validator
 ```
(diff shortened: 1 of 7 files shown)


icebob:
Nice PR, thanks. Would be good to add a strict property to the root object too (e.g. $$strict: true). What do you think?

mbaertschi:
@icebob I added the `$$strict: true` option on schema root level and enhanced the README.md accordingly

icebob:
Could you resolve merge conflicts?

icebob:
Released https://github.com/icebob/fastest-validator/releases/tag/v0.6.14

PR #50, opened 2018-11-26
Chore: Develop with ESM files

lukeed:
* Fixes `package.json` entry files:
    * add `module` entry for ESM consumers
    * fix: point `main` entry to CJS file
    * add `unpkg` entry for browser/UMD consumers
    * trims downloadable `files` list
        _no reason to have `lib` there – I'd also remove `examples` tbh._

* Updates build tools (CC)

* Replace `require` with `import` statements
     * No code changes
     * Removed `rollup-plugin-commonjs`

* Added Babel@7 configuration for Jest/env=test only
    * Still using Buble for `dist/*` contents
    * This is a lot more devDeps to download, but reflects the request made

Disabling `Object.freeze` was due to the fact that the `this.rules` were assigned a fresh clone _and_ possibly mutated. Freezing the object had no use for us here.

Additionally, I would remove Buble from the CJS/ESM output. Since both files are targeted at Node, there's not really any reason to have Buble running here.

---

Overall, I have the CJS file (`dist/index.js`) going from **5.14 kB** to **4.88 kB**~!

---

_Closes #49_

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,9 @@
+{
+	"env": {
+	    "test": {
+	      "presets": [
+	        "@babel/preset-env"
+	      ]
+	    }
+	}
+}
--- a/index.js
+++ b/index.js
@@ -1,1 +1,0 @@
-module.exports = require("./lib/validator.js");
--- a/lib/helpers/deep-extend.js
+++ b/lib/helpers/deep-extend.js
@@ -1,6 +1,4 @@
-"use strict";
-
-function deepExtend(destination, source) {
+export default function deepExtend(destination, source) {
 	for (let property in source) {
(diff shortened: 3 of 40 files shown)


lukeed:
So only Node 6.x is failing because the new Closure plugin relies on `async`. If it were me, I would choose to change the Travis matrix. The old/deprecated Closure plugin that you were using didn't successfully run on either of my machines.

icebob:
Thanks, I will check it...

erfanium:
@icebob What is the status of this? may better to convert it to draft

icebob:
I think after 1.0.0 it's outdated. Sorry @lukeed, maybe in the future.

PR #52, opened 2018-12-18
sittingbool/adding typedef

sittingbool:
adding typescript definitions file

tested in my own project using the examples as shown in readme.


sittingbool:
I believe that error is not about my changes... are you fixing it?

PR #56, opened 2019-02-05
Some code fixes and improvements

webdevium:
Some code fixes:
- made code improvements
- added `.editorconfig`
- added pre-commit for `eslint --fix`


lib/rules/email.js
```diff
@@ -1,18 +1,17 @@
...
 const BASIC_PATTERN = /^\S+@\S+\.\S+$/;
 
 module.exports = function checkEmail(value, schema) {
 	if (typeof value !== "string") {
 		return this.makeError("string");
 	}
 
-	let pattern;
-	if (schema.mode == "precise")
+	let pattern = BASIC_PATTERN;
```
akazakou: Why we should create additional variable here instead of using `PRECISE_PATTERN` constant?

Issue #28, opened 2018-06-24
Add option to set custom message error in configuration


guumo:
I think it's very important customize error for a individual configuration field, pass the message key. For example:

`first_name: { type: "string", min: 3, max: 25, message: "What is your name?" },`

ispyinternet:
Something I wanted too, I hacked in a solution that works for simple use, not sure what would be required for more complex cases like arrays, or multiple validators. 

````
const schema = {
    id: { type: "number", positive: true, integer: true , messages: { required: "id is a required field" }},
    name: { type: "string", min: 3, max: 255, messages: { stringMin: "Your name is too short" } },
    status: "boolean" // short-hand def
};
 
console.log(v.validate({ name: "J", status: true }, schema));

// output
[ { type: 'required',
    expected: undefined,
    actual: undefined,
    field: 'id',
    message: 'id is a required field' },
  { type: 'stringMin',
    expected: 3,
    actual: 1,
    field: 'name',
    message: 'Your name is too short' } ]
````

My fork is at https://github.com/ispyinternet/fastest-validator

Not a confident contributor so, if the author is up for this functionality perhaps they could implement it more elegantly / robustly?

intech:
Diff: https://github.com/icebob/fastest-validator/compare/master...ispyinternet:master

ispyinternet:
I've updated to the new code and opened a pull request.

icebob:
Released https://github.com/icebob/fastest-validator/releases/tag/v0.6.14

PR #57, opened 2019-02-05
Add feature to allow personalised error messages per field.

ispyinternet:
I've added tests and updated the readme if you are happy to include this feature.

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -36,7 +36,10 @@ function Validator(opts) {
 		deepExtend(this.opts, opts);
 
 	this.messages = this.opts.messages;
-
+	
+	// place to store custom messages
+	this.customMessages = {};
+	
 	// Load rules
 	this.rules = loadRules();
 }
@@ -153,6 +156,14 @@ Validator.prototype._processRule = function(rule, name, iterate) {
 		});
 	}
 
+	// Custom Messages - only supports simple Object rule atm.
+	if(rule.messages) {
+		if (typeof this.customMessages[name] == "undefined" ) this.customMessages[name] = {};
+		Object.keys(rule.messages).forEach( (msgKey) => {
+			this.customMessages[name][msgKey] = rule.messages[msgKey];
+		});
+	}
+	
 	return checks;
 };
 
@@ -290,7 +301,9 @@ Validator.prototype.makeError = function(type, expected, actual) {
  * @param {Object} err Validation error object
  */
 Validator.prototype.resolveMessage = function(err) {
-	let msg = this.messages[err.type];
+    let msg = this.customMessages[err.field] && this.customMessages[err.field][err.type] ? 
+			this.customMessages[err.field][err.type] : this.messages[err.type];
+
 	if (msg != null) {
 		const expected = err.expected != null ? err.expected : "";
 		const actual = err.actual != null ? err.actual : "";
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -65,6 +65,8 @@ function Validator(opts) {
 
 	this.messages = this.opts.messages;
 
+	this.messageKeys = Object.keys(this.messages);
+
 	// Load rules
 	this.rules = loadRules();
 	this.cache = new Map();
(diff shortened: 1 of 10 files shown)


ispyinternet:
@icebob I've tried to adapt. Please review.

lib/validator.js
```diff
@@ -88,220 +88,258 @@ Validator.prototype.validate = function(obj, schema) {
...
-		dataFunction = this.checkSchemaArray;
-	}
+    if (!messages["required"])
+	messages["required"] = this.messages["required"];
+    
+    if (!messages[schemaType.type])
+	messages[schemaType.type] = this.messages[schemaType.type];
+    
+    Object.keys(schemaType).forEach(rule => {
+	if (rule == "type" || rule == "items" || rule == "props" || rule == "messages") return;
```
icebob: Could you explain what is this code?

ispyinternet: In order to remove check from run time we have to prepare all possible message types at compile time. These include:

‘required’
‘string’ | ‘number’ | ‘array’ etc (i.e type)

Then we analyse the additional rules to see what other messages need to be added.

ispyinternet: I've just changed the if statement to use `key in obj` style rather than truthy test

icebob:
Good job, the performance has been improved! On my PC 5.3M ops/sec. The master branch is 5.2M :+1:

icebob:
please don't force push commit because i'm trying to linting the codes.

lib/validator.js
```diff
@@ -167,20 +170,49 @@ Validator.prototype.compileSchemaType = function(schemaType) {
...
+	if (!("required" in messages))
+		messages["required"] = this.messages["required"];
+
+	if (!(schemaType.type in messages))
+		messages[schemaType.type] = this.messages[schemaType.type];
+
+	Object.keys(schemaType).forEach(rule => {
+		if (rule == "type" || rule == "items" || rule == "props" || rule == "messages") return;
+
+		let key = schemaType.type.toLowerCase() + rule[0].toUpperCase() + rule.substring(1);
```
icebob: My problem is that it is not guaranteed that you can generate message key from type & prop name (e.g `enumValue`)

ispyinternet: oooh.. almost.

Any suggestions?

icebob: I don't understand exactly this parts, but can you forEach keys in the original `messages`? E.g. `Object.keys(this.messages).forEach(key => messages[key] = schemaType.messages[key] || this.messages[key]`?

ispyinternet: shouldn't that key just be 'enum'?

ispyinternet: and we add stringEnum etc

icebob: it can be, just it caused a breaking change.

ispyinternet: that would work, it would mean passing all of the messages along with every rule. Not sure how significant that is and how you feel about that?

icebob: Yes, but I don't think it can cause any problem. Just some more bytes in the memory.

ispyinternet: ok since my linting is bad do you want to make update?

ispyinternet: or do I need to? This is my first pull request!

icebob: I added editorconfig to this PR and fix lintings. Pull these changes and check your editor use linting based on editorconfig. E.g. your commit convert tabs to spaces but the repo uses tabs in source files.

ispyinternet:
All tests are passing, not sure if you want to review performance or do any further sanity checks. Some unexpected changes to review.

icebob:
The `compile` is very bad. It comes with the latest commit. I think due to `Object.keys` because it is a very expensive function.

ispyinternet:
Also for every rule we are now iterating over all the default messages. Previously only iterates if user supplied. Perhaps an overwriting merge of the two objects. I’ll make the update. Is there a command to run so I can get the metrics to report back?

icebob:
I've added a condition, so only handle custom messages if there is `messages` property in schema.

icebob:
Now it's much better.
It doesn't affect the performance without messages, and with it now is ~70k

√ compile & validate                                119,438 rps
√ compile & validate with custom messages            69,858 rps
√ validate with pre-compiled schema               5,220,102 rps
√ validate with wrong obj                           660,502 rps

ispyinternet:
Nice! Why did you remove Object.keys(this.messages) from constructor? Now it is called everytime a message property is encountered - negligible admittedly!?

Also did you see the slight changes to `lib/rules/string.js` and `lib/rules/email.js`?

icebob:
Because there is not relevant difference. With Object.keys ~68k, with static array ~70k. But I'm checking again.

In string & email files I made some linting in regexp.

icebob:
@ispyinternet I put back the key array. It gives +10%. So now I think it's ready & performant. Could you add this new feature to the README too with examples?

ispyinternet:
I already updated the README its the last section

icebob:
Really, sorry :)

icebob:
Thank you for your work.

ispyinternet:
Thanks for your help too!

ispyinternet:
you can close #28

Issue #64, opened 2019-02-16
Typedefs are broken


anri-asaturov:
Typedefs are so broken, guys, please consider disabling them until sensible PR is made.

some examples: 
- missing 'empty'
- 'items' can be an object

I'm sure I will find more if i continue looking at api.

akazakou:
Also not working multiple validators for `validator.compile` method

couellet:
Definition file looks good in the repo though, but somehow npm package is out of sync :/

icebob:
Published in v0.6.16

couellet:
My bad, the defs file wasn't correct after all. I just submit a pull request fixing it.

v1tail:
https://github.com/icebob/fastest-validator/blob/master/index.d.ts doesn't work with the new release (0.6.16). TypeScript signals an error "An export assignment cannot be used in a module with other exported elements."

akazakou:
@v1tail Yes. My bad. I fixed it in icebob/fastest-validator#68 PR but it still opened

PR #65, opened 2019-02-21
Fix wrong TypeScript types

akazakou:
Changeset to #64 issue:
* Exporting defined types
* Remove declare construction that is not required
* Fix `compile` to allow an array of validation objects as an input parameter
* Update `check` function, it returns only `true` or error description


akazakou:
Someone can review it?

icebob:
@anri-asaturov ?

PR #67, opened 2019-03-06
Fix $$strict property compilation

mbaertschi:
The `$$strict` property was not evaluated correctly for nested objects. This PR fixes the problem.


mbaertschi:
> Great, thank you. Please add relevant tests, too.

Done

PR #74, opened 2019-05-12
Resolve #15: Convert original boolean/number values if set flag convert true

### LINKED ISSUE #15: convert true to boolean don't actual converts the value.

alexeymarunin:
I tried to resolve #15 

```javascript
const schema = {
	id: { type: "number", positive: true, integer: true, convert: true },
	name: { type: "string", min: 3, max: 255 },
	status: { type: "boolean", convert: true }
};

const data = { id: "5", name: "John", status: "1" };
console.log(v.validate(data, schema), data);
// Returns: true { id: 5, name: 'John', status: true }
```


alexeymarunin:
@icebob is it any problems to submit the request? 
This feature is important for our project

icebob:
Yes, because it is a breaking change. Because it modifies the original object.

icebob:
https://github.com/icebob/fastest-validator/pull/94

PR #85, opened 2019-09-29
add rules mac and luhn

intech:
@icebob I want a new version release =)


icebob:
I'm going to release tonight.

PR #94, opened 2019-11-10
New rewritten version 1.0.0

icebob:
The full library has been rewritten. It uses code generators in order to be much faster.

## Breaking changes
This new version contains several breaking changes.

### Rule logic changed
The rule codes have been rewritten to code generator functions. Therefore if you use custom validators, you should rewrite them after upgrading.

### Convert values
The `number`, `boolean` and `date` rules have a `convert: true` property. In the previous version it doesn't modify the value in the checked object, just converted the value to the rules. In the version 1.0 this property converts the values in the checked object, as well.

### Sanitizations
The sanitization function is implemented. There are several rules which contains sanitizers. **Please note, the sanitizers change the original checked object values.**

| Rule | Property | Description |
| ---- | -------- | ----------- |
`boolean` | `convert` | Convert the value to a boolean.
`number` | `convert` | Convert the value to a number.
`date` | `convert` | Convert the value to a date.
`string` | `trim` | Trim the value.
`string` | `trimLeft` | Left trim the value.
`string` | `trimRight` | Right trim the value.
`string` | `lowercase` | Lowercase the value.
`string` | `uppercase` | Uppercase the value.
`string` | `localeLowercase` | Lowercase the value with `String.toLocaleLowerCase`.
`string` | `localeUppercase` | Uppercase the value with `String.toLocaleUpperCase`.
`string` | `padStart` | Left padding the value.
`string` | `padEnd` | Right padding the value.
`string` | `convert` | Convert the value to a string.
`email` | `normalize` | Trim & lowercase the value.
`forbidden` | `remove` | Remove the forbidden field.
`object` | `strict: "remove"` | Remove additional properties in the object.
`*` | `default` | Use this default value if the value is `null` or `undefined`.

### Root element validation
Basically the validator expects that you want to validate a Javascript object. If you want others, you can define the root level schema, as well. In this case set the `$$root: true` property.

**Example to validate a `string` variable instead of `object`**
```js
const schema = {
    $$root: true,
    type: "string", 
    min: 3, 
    max: 6
};

v.validate("John", schema); // Valid
v.validate("Al", schema); // Fail, too short.
```

### Enhanced shorthand types
You can use string-based shorthand validation definitions in the schema with properties.

```js
{
    password: "string|min:6",
    age: "number|optional|integer|positive|min:0|max:99",

    retry: ["number|integer|min:0", "boolean"] // multiple types
}
```

### New `equal` rule
It checks the value equal (`==`) to a static value or another property. The `strict` property uses `===` to check values.

**Example with static value**:
```js
const schema = {
    agreeTerms: { type: "equal", value: true, strict: true } // strict means `===`
}

v.validate({ agreeTerms: true }, schema); // Valid
v.validate({ agreeTerms: false }, schema); // Fail
```

**Example with other field**:
```js
const schema = {
    password: { type: "string", min: 6 },
    confirmPassword: { type: "equal", field: "password" }
}

v.validate({ password: "123456", confirmPassword: "123456" }, schema); // Valid
v.validate({ password: "123456", confirmPassword: "pass1234" }, schema); // Fail
```

### `properties` in object rule
You can use the `properties` property besides the `props` property in the object rule.


icebob:
@akazakou could you help me with the typescript definition & test?
I would like to create a ts test but I've got error message:
[Source](https://github.com/icebob/fastest-validator/blob/next/test/typescript/full/index.ts)
Error:
image

alexeymarunin:
@icebob great work!
Are you planning to integrate 1.x.x version with Moleculer?

icebob:
Yes, Moleculer v0.14 will use it.

akazakou:
@icebob I need some time to review updated code and rewrite TypeScript definitions. It will take ~2-3 days.

icebob:
@akazakou Ok. I've tried to update & add changes but I can't check that it's correct. And as I mentioned, would be good to add a typescript test as well, but I've got error message when I imported the validator.

akazakou:
@icebob do not block you, I've integrated TypeScript support for Jest unit tests. Please check my PR #95

icebob:
Thank the #95, but now I have another problem :(
I've added a simple schema but I've got errors again: https://github.com/icebob/fastest-validator/pull/94/commits/034c16937874e038e05828286c0c4d7b0488b33c
image

akazakou:
Some workaround for this: 
```typescript
/// <reference path="../../index.d.ts" /> // here we make a reference to exists module definition
import ValidatorType, { RuleNumber, ValidationSchema } from 'fastest-validator'; // here we importing type definition of default export

const Validator: typeof ValidatorType = require('../../index'); // here we importing real Validator Constructor

describe('Typescript Definitions', () => {
    it('should compile validator', async () => {
        const v: ValidatorType = new Validator();
        const compiled = v.compile({
			id: { type: "number", min: 1, integer: true } as RuleNumber
		} as ValidationSchema);

        expect(compiled).toBeInstanceOf(Function);

		const res = compiled({
			id: 2
		});

		expect(res).toBe(true);
    });
});
```

I'm investigating, why it stops working...

akazakou:
@icebob I've created a PR #96 to verify that all exists unit tests (except helpers) will work correctly under TypeScript control

rfranco:
@icebob could you publish a version as _next_?
I'd like to test it.

icebob:
You can try it with
`npm i fastest-validator@next`
or in browser
https://unpkg.com/fastest-validator@next

akazakou:
Great job @icebob

rfranco:
It's work `@next`.

I'll test it, also may I update the #50 ?
Because I'd like to use this in server and browser side.

PR #95, opened 2019-11-11
Integrate Jest tests for TypeScript definitions

akazakou:
### Description
Here I've done a kludge to import type definitions without rewriting exists type definitions. 

I've added [Triple-Slash Directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) to allow import of `fastest-validator` types. 
```
/// <reference path="../../../index.d.ts" />
```

Next, I've imported default export from `fastest-validator` definition package 
```
import ValidatorType from 'fastest-validator';
```

And the last one I've imported from javascript file default export and assigned to this imported variable, exported from 'fastest-validator' definition type.

```
const Validator: typeof ValidatorType = require('../../../index');
```

Now constant `Validator` have a type named as `ValidatorType` that actually `Validator` class.

### Changelog
* Added @types/jest, rollup-plugin-copy and ts-jest dev dependencies
* Added copying index.d.ts file into dist folder
* Integrated TypeScript test via ts-jest support

Now you no need dedicated `test:ts` npm job. You will be enough to run `test` job and it will test Javascript and Typescript files at the same time.


icebob:
Awesome job @akazakou! Thank you very much and thank you for your quick job!

PR #96, opened 2019-11-12
Adopt exists unit tests for using in TypeScript

akazakou:
### Changelog
* Copy and adopt exists unit tests for a run under TypeScript control


index.d.ts
```diff
@@ -679,7 +680,7 @@ declare module 'fastest-validator' {
 	/**
 	 * Description of validation rule definition for a some property
 	 */
-	type ValidationRule = ValidationRuleObject | ValidationRuleObject[] | ValidationRuleName | string;
+	type ValidationRule = ValidationRuleObject | ValidationRuleObject[] | ValidationRuleName | number;
```
icebob: Why `number`?

akazakou: Nice catch. I've tested negative cases for `v.compile(1)` to throw errors

icebob: Travis failed :(

akazakou: Yeah. Some last time moment fixes :(

akazakou: Now it's green

PR #104, opened 2019-12-13
add unique validation in array rule

intech:


intech:
@icebob I updated everything to the latest version.

I missed and my fork lagged behind such global code changes.

lib/rules/array.js
```diff
@@ -54,6 +54,14 @@ module.exports = function({ schema, messages }, path, context) {
 		`);
 	}
 
+	if (schema.unique === true) {
+		src.push(`
+			if(len > Array.from(new Set(value)).length) {
```
icebob: The `Set` has a `size` property, no need to convert it back the array for length. Please update it.

PR #111, opened 2020-02-08
Fix string with pattern where regular expression contains a double quote

FranzZemen:
For the string rule with pattern, if the regular expression contains a double quote an invalid token error occurs in the message field.  This is because the RegExp.toString does not escape the internal double quote and the construction of the message concatenates double quotes to this value.  

You can test this with sample regular expression /^[a-z0-9 .\-'?!":;\\/,_]+$/i

In this pull request, when constructing the message I replace the double quote with an escaped version, but due to the internal processing I have to also escape the escape character (\\\").

Jest reports 100% success for all tests plus the additional test I added.

### DIFF
--- a/lib/rules/string.js
+++ b/lib/rules/string.js
@@ -137,7 +137,7 @@ module.exports = function checkString({ schema, messages }, path, context) {
 
 		src.push(`
 			if (!${pattern.toString()}.test(value))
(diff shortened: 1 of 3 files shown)


icebob:
Released in 1.0.2

PR #118, opened 2020-03-11
Implement Aliases

erfanium:
Sometime I need use a same validation rule in some places, aliases (or whatever name is better) are great for this case.
For example, I define an alias for my username validation rule:

```js
v.alias('username', {
   type: 'string',
   min: 4,
   max: 30
   regex: ...
  ....
})
``` 
Then, I can use this alias everywhere I want: 

```js
v.compile({
   firstName: 'string',
   email: 'email',
   username: 'username'  // using alias 
}) 
``` 
also, extending aliases!

```js
v.compile({
   firstName: 'string',
   email: 'email',
   username: {
      type: 'username',
      optional: true,
      max: 60
   }
}) 
```

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -49,6 +49,7 @@ class Validator {
 
 		// Load rules
 		this.rules = loadRules();
+		this.aliases = {};
 		this.cache = new Map();
 	}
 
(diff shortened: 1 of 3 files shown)


erfanium:
@icebob 
Fixed. 
To avoid this problem, it's better to use Prettier , etc...

lib/validator.js
```diff
@@ -281,6 +282,15 @@ class Validator {
 				schema.optional = true;
 		}
 
+		const alias = this.aliases[schema.type];
+
+		if (alias) {
+			schema.type = alias.type;
```
icebob: Why set `schema.type` if in the next line you remove it?

lib/validator.js
```diff
@@ -281,6 +282,15 @@ class Validator {
...
+			delete schema.type;
+			Object.assign(alias, schema);
```
icebob: It doesn't correct because it overwrites the original alias. So if you use it multiple times it won't be the same.
Use `Object.assign({}, alias, schema)` to create a new instance.

icebob:
Thank you.

erfanium:
Just don't forget README.md
I think it's better if you change it. I'm not very good at English :-)

PR #119, opened 2020-03-16
Custom validation for built-in rules

erfanium:
Related to #100 
Customizing built-in rules, without creating new rule. 
For example:
```js
check = v.compile({
   num: {
      type: "number",
      min: 10,
      max: 15,
      integer: true,
      custom (value, schema) {
         if (value % 2 !== 0) return [{ type: "evenNumber", actual: value }];
      }
   }
});
```

### DIFF
--- a/examples/full.js
+++ b/examples/full.js
@@ -37,6 +37,7 @@ const schema = {
 	apikey: "forbidden",
 	uuidv4: { type: "uuid", version: 4 },
 	uuid: "uuid",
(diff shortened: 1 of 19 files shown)


erfanium:
Why is this happening?
I have no failed test on my PC.
I think the problem is with Travis CI cache. can I retry?

erfanium:
@icebob 
Any opinion? Proposal? Criticism? :-)

icebob:
The idea is good.
But I should check the code & check the performance of PR. If it's ok I will merge it.

PR #120, opened 2020-03-21
Rules default settings  

erfanium:
to change default rules behavior.
For example: object strict

```js
‍‍‍const v = new Validator({
   defaults: {
      object: {
         strict: 'remove'
      }
   }
})
```

### DIFF
--- a/test/helpers/deep-extend.spec.js
+++ b/test/helpers/deep-extend.spec.js
@@ -39,3 +39,21 @@ describe("deepExtend", () => {
 		});
 	});
 });
+
+describe("Test merge options", () => {
+	it("should consider overrideDist option", () => {
+		const result = deepExtend({
+			b: 5,
+			c: 6
+		}, {
+			b: 10,
+			e: "Hello"
+		}, { overrideDest: false });
+
+		expect(result).toEqual({
+			b: 5,
+			c: 6,
+			e: "Hello"
+		});
+	});
+});
--- a/test/helpers/deep-extend.spec.js
+++ b/test/helpers/deep-extend.spec.js
@@ -41,14 +41,14 @@ describe("deepExtend", () => {
 });
 
 describe("Test merge options", () => {
-	it("should consider overrideDist option", () => {
+	it("should consider skipIfExist option and not overwrite the existing properties", () => {
 		const result = deepExtend({
 			b: 5,
 			c: 6
 		}, {
 			b: 10,
 			e: "Hello"
-		}, { overrideDest: false });
+		}, { skipIfExist: true });
 
 		expect(result).toEqual({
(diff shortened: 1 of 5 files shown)


test/helpers/deep-extend.spec.js
```diff
@@ -39,3 +39,21 @@ describe("deepExtend", () => {
...
+	it("should consider overrideDist option", () => {
+		const result = deepExtend({
+			b: 5,
+			c: 6
+		}, {
+			b: 10,
+			e: "Hello"
+		}, { overrideDest: false });
+
+		expect(result).toEqual({
```
icebob: Please check that it really doesn't modify the source object, as well.

erfanium: That's exactly why it was.
because rule default settings must not override user schema property that already exist. it should only add new property/key to object - not overriding

erfanium: and `Object.assign({}, ....)` doesn't work here. because it creates a new Object every time and disrupt the cacher.

icebob: I mean, checks the logic that it doesn't modify the original object, like: 

```js
	it("should consider overrideDist option", () => {
		const obj = {
			b: 5,
			c: 6
		};
		const result = deepExtend(obj, {
			b: 10,
			e: "Hello"
		}, { overrideDest: false });

		expect(result).toEqual({
			b: 5,
			c: 6,
			e: "Hello"
		});
		expect(obj).toEqual({
			b: 5,
			c: 6
		});        
	});
```

erfanium: It's ok, because `overrideDest` is false 
The key "b = 10" from source object is already exist in destination object and `deepExtand` function must not override this key. only add **new** key

icebob: I understand what you said. My problem is that, if you didn't check that it doesn't touch the destination, and somebody changes this function in the future, we won't able to detect that this override logic still works well.

icebob: By the way, just copy my previous code parts to the test & I'm fine to merge it :)

erfanium: Your previous test code will be fail. cause `deepExtend` function always return destination object as result.
This function always modifies destination (with any options). The only difference between options is the logic of extending.
I may have chosen a bad name for this option that you were confused about.

icebob: Hmm, Ok, I will check it again.

icebob: Ok, now I see what you tried to say :) I've renamed the option & changed the logic to the opposite.

PR #122, opened 2020-03-30
Type defs for custom validation, fixes.

FFKL:
Hello! I've improve some type definitions.

- CompilationRule, Context, CheckerFunction, CompilationFunction for custom rule definitions (directly in schema or by `add` method).
- RuleCustomInline upd: I think, generic isn't necesssary. Input type for validation - `unknown`.
- ValidationSchema fix: ability to use of root-level schema fields. For example - `min: 1`
- ValidationError fix: `type` changed to message name, `message` - string for custom message
- `compile`, `validate` methods fix: now it can be `any` type - not only object.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -403,18 +403,15 @@ declare module 'fastest-validator' {
 	 * Validation schema definition for custom inline validator
 	 * @see https://github.com/icebob/fastest-validator#custom-validator
 	 */
-	interface RuleCustomInline<T = any> extends RuleCustom {
+	interface RuleCustomInline extends RuleCustom {
 		/**
 		 * Name of built-in validator
 		 */
 		type: 'custom';
 		/**
-		 *
-		 * @param {{any}} value Value that should be validated
-		 * @param {ValidationRuleObject} schema Validation schema that describes current custom validator
-		 * @return {{true} | ValidationError[]} true if result is valid or array of validation error messages
+		 * Custom checker function
 		 */
-		check: (value: T, schema: ValidationRuleObject, path: string, parent?: object, context?: any) => true | ValidationError[];
+		check: CheckerFunction;
 	}
 
 	/**
@@ -443,12 +440,9 @@ declare module 'fastest-validator' {
 		default?: any;
 
 		/**
-		 *
-		 * @param {{any}} value Value that should be validated
-		 * @param {ValidationRuleObject} schema Validation schema that describes current custom validator
-		 * @return {{true} | ValidationError[]} true if result is valid or array of validation error messages
+		 * Custom checker function
 		 */
-		custom?: (value: unknown, schema: ValidationRuleObject, path: string, parent?: object, context?: any) => true | ValidationError[];
+		custom?: CheckerFunction;
 
 		/**
 		 * You can define any additional options for custom validators
@@ -696,17 +690,22 @@ declare module 'fastest-validator' {
 	interface ValidationSchema {
 		/**
 		 * Object properties which are not specified on the schema are ignored by default.
-		 * If you set the $$strict option to true any aditional properties will result in an strictObject error.
+		 * If you set the $$strict option to true any additional properties will result in an strictObject error.
 		 * @default false
 		 */
 		$$strict?: boolean;
 
+		/**
+		 * Basically the validator expects that you want to validate a Javascript object.
+		 * If you want others, you can define the root level schema.
+		 * @default false
+		 */
 		$$root?: boolean;
 
 		/**
 		 * List of validation rules for each defined field
 		 */
-		[key: string]: ValidationRule | boolean | undefined;
+		[key: string]: ValidationRule | undefined | any;
 	}
 
 	/**
@@ -716,7 +715,7 @@ declare module 'fastest-validator' {
 		/**
 		 * Name of validation rule that generates this message
 		 */
-		type: ValidationRuleName;
+		type: keyof BuiltInMessages | string;
 		/**
 		 * Field that catch validation error
 		 */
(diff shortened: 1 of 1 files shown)


FFKL:
@icebob in source code i found `sanitized` property for validation object - `{ sanitized: boolean, source: string}`. What is it used for?

icebob:
sanitized & source are used internally only.

icebob:
@akazakou could you review, plz?

index.d.ts
```diff
@@ -403,18 +403,15 @@ declare module 'fastest-validator' {
 	 * Validation schema definition for custom inline validator
 	 * @see https://github.com/icebob/fastest-validator#custom-validator
 	 */
-	interface RuleCustomInline<T = any> extends RuleCustom {
```
akazakou: Not sure that we should remove generic form here. It should help detect errors in the pre-compile stage of a project.

index.d.ts
```diff
@@ -746,6 +745,25 @@ declare module 'fastest-validator' {
...
+	}
+
+	interface Context {
+		index: number;
+		rules: ValidationRuleObject[];
+		fn: Function[];
+		customs: { [ruleName: string]: { schema: RuleCustom, messages: MessagesType } }
+	}
+
+	type CheckerFunction = (value: unknown, schema: ValidationSchema, path: string, parent: object | null, context: Context) => true | ValidationError[];
```
akazakou: `CheckerFunction` should have a generic type for `value` parameter. 
```suggestion
	type CheckerFunction<T = unknown> = (value: T, schema: ValidationSchema, path: string, parent: object | null, context: Context) => true | ValidationError[];
```

index.d.ts
```diff
@@ -792,22 +810,22 @@ declare module 'fastest-validator' {
 		 * @param {any=} opts.actual
 		 * @param {MessagesType} opts.messages
 		 */
-		makeError(opts: { type: keyof MessagesType, field?: string, expected?: any, actual?: any, messages: MessagesType }): ValidationError;
+		makeError(opts: { type: keyof MessagesType, field?: string, expected?: any, actual?: any, messages: MessagesType }): string;
```
akazakou: Are you sure that the method should keep in the current definition? It looks like an internal method and shouldn't be exposed outside of current definitions

index.d.ts
```diff
@@ -746,6 +745,25 @@ declare module 'fastest-validator' {
 		messages?: MessagesType,
 	};
 
+	interface CompilationRule {
```
akazakou: Please add to the current section comments for a better understanding of what it doing.

akazakou: Because right now it looks like that `CompilationRule` required for `CompilationFunction`, and `CompilationFunction` required for `CompilationRule`

index.d.ts
FFKL: It is used in [custom validator](https://github.com/icebob/fastest-validator#custom-validator) creation.

akazakou: Ok. That's weird, but I'm ok with it

index.d.ts
```diff
@@ -716,15 +715,15 @@ declare module 'fastest-validator' {
 		/**
 		 * Name of validation rule that generates this message
 		 */
-		type: ValidationRuleName;
+		type: keyof BuiltInMessages | string;
```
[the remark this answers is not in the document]
FFKL: No) Its type of message (key in `BuiltInMessages`).
For example:
https://github.com/icebob/fastest-validator/blob/7f08bd57b7a5d7fbea6e725775e2a76480e7ba50/test/typescript/rules/string.spec.ts#L35

index.d.ts
FFKL: I think, generic type it's way to possible errors. `unknown` makes us to check all possible validation cases. But generic type promise correct input type - it's wrong. Originally input value is unknown.

index.d.ts
FFKL: The same case as in the previous conversation

FFKL:
> sanitized & source are used internally only.

Maybe `sanitized` but not `source` - https://github.com/icebob/fastest-validator#custom-validator

index.d.ts
FFKL: - `CompilationFunction` - is used for [custom validator creation.](https://github.com/icebob/fastest-validator#custom-validator). 
- `CompilationRule` - `{ messages, schema }` in this example. I've added fields according to source code. It also includes link to `CompilationFunction`

FFKL: I forgot to change `add` method signature. Already fixed.

index.d.ts
akazakou: Ok. But we should keep in mind that is a breaking change and will make users update code after the update

index.d.ts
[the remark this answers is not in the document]
akazakou: Ok. Got it

index.d.ts
icebob: Can we solve that it doesn't cause breaking change? I don't plan to release a new major version.

FFKL: @icebob Generic types for RuleCustomInline and CheckerFunction was returned

index.d.ts
akazakou: Guess it can be resolved by that comment https://github.com/icebob/fastest-validator/pull/122#discussion_r401724595

index.d.ts
icebob: Thanks!

icebob:
@akazakou @FFKL so can I merge?

FFKL:
> @akazakou @FFKL so can I merge?

I think - yes)

PR #128, opened 2020-04-16
deep-extend: fix object type checking bug

erfanium:
`typeof o === "object"` doesn't always return expected result (e.g: with array, null, regexp, ....).

### DIFF
--- a/lib/helpers/deep-extend.js
+++ b/lib/helpers/deep-extend.js
@@ -1,12 +1,13 @@
 "use strict";
 
-function isObject(v) {
-	return v != null && v.constructor.name === "Object";
+function isObjectHasKeys(v) {
+	if (typeof v !== "object" || v == null) return false;
+	return Object.keys(v).length > 0;
 }
 
 function deepExtend(destination, source, options = {}) {
 	for (let property in source) {
-		if (isObject(source[property]) && source[property] !== null) {
+		if (isObjectHasKeys(source[property])) {
 			destination[property] = destination[property] || {};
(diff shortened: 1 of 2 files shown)


lib/helpers/deep-extend.js
```diff
@@ -1,12 +1,16 @@
 "use strict";
 
+function isObject(v) {
+	return v != null && v.constructor.name === "Object";
```
icebob, line 4: It can throw error if `v` is created as `Object.create(null)`
```
TypeError: Cannot read property 'name' of undefined
```

erfanium: Do you have any code suggestion?
I checked the `lodash` source code for its `isPlainObject` method, but it has lot of codes

icebob: Yeah, I've also checked, it's very big and I think that's very slow.
What was the exact problem with the original code? Maybe better if we extend it

erfanium: > and I think that's very slow.

not a big problem, because It will only run once at compile

icebob: In this case, just check that there is `v.constructor`. If yes, check the name, if not check the typeof.

erfanium: what about this?
```js
function isObjectHasKey(v) {
	if (v == null || typeof v !== "object") return false;
	return Object.keys(v).length > 0;
}
```

erfanium:
@icebob 
I think this code is better

PR #132, opened 2020-04-28
HexString support

erfanium:


test/rules/string.spec.js
```diff
@@ -141,6 +141,24 @@ describe("Test rule: string", () => {
 
 	});
 
+	it("check hex string", () => {
+		const check = v.compile({ $$root: true, type: "string", hex: true});
+		const message = "The '' field must be a hex string.";
+
+		expect(check("ABCD")).toEqual([{type: "stringHex", actual: "ABCD", message }]);
```
icebob, line 148: Why it doesn't accept hexstring with uppercase characters?

erfanium: accepting both types can cause unwanted problems for developers.
Developers usually choose lowercase, but with set `lowercase: true` in rule option, it will converts uppercase to lowercase and as a result: accepts both types.

I can revert this, If you think it doesn't make sense

erfanium: Another way is that specifying the exact type in rule option. like: 
```js
{
   type: "string",
   hex: "lowercase" // or "uppercase" ,or "both" 
}

but i think this is over engineering!
```

icebob: If I check the definition, it says: 
>In mathematics and computing, hexadecimal (also base 16, or hex) is a positional system that represents numbers using a base of 16. Unlike the common way of representing numbers with ten symbols, it uses sixteen distinct symbols, most often the symbols "0"–"9" to represent values zero to nine, and "A"–"F" (or alternatively "a"–"f") to represent values ten to fifteen.

So I think we should accept both lowercase & uppercase characters as well. And the developers can change the case with `lowercase: true` or `uppercase: true`.

PR #133, opened 2020-05-02
New checker function API + some other changes

erfanium:
1. BREAKING CHANGE: new checker function API
In current checker function, we can't modify the value, this new API solves this problem. 
example:
```js
v.compile({
   id: {
      type: "custom",
      check(value, errors) {
          try {
             return new ObjectID(value)
          } catch(e) {
             errors.push({type: "objectID"})
          }
      }
   }
})
```

2. Validator.makeCustomValidator method:  it's better than current `customValidator` source generator function that makes code harder to understand.

3. FIX: index.d.ts uses single quotes instead of doable quotes.
4. update jest, jest-cli, ts-jest to latest version.

README.md needs update, I will do that if you (@icebob) agree with these changes.


icebob:
Good job, but would be good to avoid the breaking change. Can you do it that we can switch the old & new checker function? Maybe with a prop in the rule properties and the default is the old checker function. Or with a new type `"custom2"` or something else. What do you think?

erfanium:
> Maybe with a prop in the rule properties.

I think it's better to put this in FastestValidator class constructor options:

```js
const v = new FastestValidator({
    useNewCheckerFunction: true // default: false
})
```
and show a deprecated-warning message, if they use old one

icebob:
Ok, just use the `useNewCustomCheckerFunction` option name

erfanium:
@icebob 
Can I finalize this?

test/typescript/rules/custom.spec.ts
```diff
@@ -14,23 +63,26 @@ describe('TypeScript Definitions', () => {
...
         it('should handle returned errors', () => {
-            const checker = jest.fn(function(value, schema, field) {
-                return [{ type: 'myError', field, expected: 3, actual: 4 }];
-            });
+            const fn: CheckerFunction<string> = function (value, errors, schema, field) {
+                errors.push({ type: 'myError', field, expected: 3, actual: 4 });
+                return value
+            }
+
+            const checker = jest.fn(fn as any);
```
erfanium, line 85: I don't know why it makes TypeError if I remove `as any`
I think it's problem from Jest Typing. because types are OK without using `jest.fn`

lib/validator.js
```diff
@@ -47,6 +47,8 @@ class Validator {
 		if (opts) {
 			deepExtend(this.opts, opts);
 			if (opts.defaults) deepExtend(this.defaults, opts.defaults);
+			// eslint-disable-next-line no-console
+			if (!this.opts.useNewCustomCheckerFunction) console.warn("DeprecationWarning: set 'useNewCustomCheckerFunction = true' in FastestValidator constructor"); // todo: add link
```
icebob, line 51: This mode is very aggressive. It means, all existing users will get this warning message, either don't use custom rules. And in Moleculer you can't add Validator options easily, so they will always get this warning. 
I think no need this warning because old-style will work in v1.x version. After v2.x will be this the default but we will denote It in the changelog. So plz, remove this warning.

lib/rules/email.js
```diff
@@ -26,11 +26,11 @@ module.exports = function({ schema, messages, customValidation }, path) {
...
 
 	src.push(`
-		if (!${pattern.toString()}.test(value))
+		if (!${pattern.toString()}.test(value)) {
 			${this.makeError({ type: "email",  actual: "value", messages })}
+		}
 
-		${customValidation("value")}
-
+		${this.makeCustomValidator({ path, schema, messages, context })}
```
icebob, line 33: As I see, all rules code contains the same source code. Can we move it into the common parts in validator.js from here? Similar to the required/optional checks which are same in all rules.

lib/validator.js
erfanium: Users can use old fastest-validator versions without worry! They will only see this message when they have updated the fastest-validator version, I've seen this pattern in many other packages, like mongodb (native driver) - express, etc.. , it's pretty common.

I have no problem with removing this, but a trigger is needed to that users update their codes

erfanium: You can use current fastest-validator version(1.3.0) for Moleculer until it reaches version 2.

lib/rules/email.js
erfanium: Yesss!

lib/validator.js
icebob: Moleculer uses `^1.0.0` so the validator can be updated automatically. So my problem is that in Moleculer you can't set this option easily.

icebob:
Could you help in README updating, as well?

PR #142, opened 2020-05-22
[Feature] Add support for minProps and maxProps in object validation

alexjab:
Hi, thanks for the good job you've been doing with `fastest-validator`!

Would you be interested in two new rules `minProps` and `maxProps` in the object validation? These rules would be the minimum required props in an object (and maximum allowed props).

These features exist in JSON Schema, under the names `minProperties` and `maxProperties`, but since `fastest-validator` seems to use the name `props`, I figured I'd use `minProps` and `maxProps` instead.

What do you think about that?

BR,

Alex

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -331,8 +331,8 @@ class Validator {
 		};
 		if (field) o.field = `"${field}"`;
 		else o.field = "field";
-		if (expected) o.expected = expected;
-		if (actual) o.actual = actual;
+		if (expected != null) o.expected = expected;
+		if (actual != null) o.actual = actual;
 
 		const s = Object.keys(o)
 			.map(key => `${key}: ${o[key]}`)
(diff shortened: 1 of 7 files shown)


erfanium:
@alexjab Great, but `index.d.ts` also needs to be updated.

alexjab:
I've updated the typings, don't hesitate if you need other fixes.

erfanium:
@alexjab LGTM, thanks

lib/validator.js
```diff
@@ -331,8 +331,8 @@ class Validator {
 		};
 		if (field) o.field = `"${field}"`;
 		else o.field = "field";
-		if (expected) o.expected = expected;
-		if (actual) o.actual = actual;
+		if (expected != null) o.expected = expected;
+		if (actual != null) o.actual = actual;
```
icebob, line 335: Nice catch 👍

PR #146, opened 2020-06-02
Added option to use generics with the `compile` method

Gamote:
I've added support for generics on the `compile` method in order to take advantage of the Typescript validation when we are defining the schema as a param.

**Usage (before):**
```ts
type LoginRequest = {
  email: string;
  password: string;
};

const v = new Validator();

type CValidationSchema<T = any> = {
  $$strict?: boolean | 'remove';
  $$root?: boolean;
} & {
  [key in keyof T]: ValidationRule | undefined | any;
}

type CValidator<T> = (value: T) => true | ValidationError[];

const typedCompiler = <T>(schema: CValidationSchema<T>): CValidator<T> =>
  v.compile(schema);

const loginRequestValidator = typedCompiler<LoginRequest>({
  email: {type: 'email'},
  password: {type: 'string', empty: false, min: 6},
});
```

**Usage (after):**
```ts
type LoginRequest = {
  email: string;
  password: string;
};

const v = new Validator();

const loginRequestValidator = v.compile<LoginRequest>({
  email: {type: 'email'},
  password: {type: 'string', empty: false, min: 6},
});
```

I've set the `T` default value to `any` so the `compile` method can still be used without generics.

The most concerning part for me, regarding this commit, is that I've changed the `ValidationSchema` from **interface** to **type**. I've checked the code and I couldn't find any possible issue with it but I'm not totally sure about all the consequence. So please let me know if you find something.

If everything it's alright, I will continue with other PRs.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -715,11 +715,11 @@ declare module "fastest-validator" {
 		| ValidationRuleObject
 		| ValidationRuleObject[]
 		| ValidationRuleName;
-
+	
 	/**
 	 * Definition for validation schema based on validation rules
 	 */
-	interface ValidationSchema {
+	type ValidationSchema<T = any> = {
 		/**
 		 * Object properties which are not specified on the schema are ignored by default.
 		 * If you set the $$strict option to true any additional properties will result in an strictObject error.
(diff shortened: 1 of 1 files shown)


erfanium:
🤔🤔🤔
Changes looks great to me. But I think extending the `CheckerFunction` It's not as complicated as you think.

```ts
interface CustomCheck<T> {
   (schema: T): true | FV.ValidationError[]
}

interface Schema {
   name: string
}

const check: CustomCheck<Schema>  = validatorBase.compile({
   name: 'string'
})
```

Issue #124, opened 2020-03-31
Feature Request: Validate array ordered


simlu:
This is important when e.g. validating GPS coordinate tuples.

Could allow array to be passed for `items` parameter in array?

Please let me know what you think about this addition! Cheers!

icebob:
yes, e.g.:
```js
{
    type: "array", items: {
        type: "array", items: {
            type: "number"
        }
    }
}
```

simlu:
Hello icebob. This is not what I meant at all.

I'm trying to get validation for `[180, 90]` valid, but `[90, 180]` invalid.

icebob:
I see. It is not available. You should do it with a custom validator only.

simlu:
@icebob Would you be ok with extending the functionality with an additional flag on the array `exact=BOOL`? After reviewing this I think that's all that would be needed here

icebob:
And what will be the logic behind `exact`? The `items` will be an array with schemas?

simlu:
We can already provide an array for items. So it would just ensure
- array length matches items length
- the entries in array validate against corresponding items entries

Schema would look as follows:

```json
{
    type: "array", exact: true, items: [
      { type: "number", min: -180, max: 180 },
      { type: "number", min: -90, max: 90 }
    ]
}
```

erfanium:
@simlu @simlu 
We can add a new type named: **Tuple**, like Typescript:
https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple

icebob:
A new rule type? Or inside `array` type?

erfanium:
@icebob new rule. like
```
{
   foo: {
      type: 'tuple',
      0: 'string',
      1: {
             type: 'number',
             min: 10 
          }
   }
}
```

icebob:
I like the second option, it follows the structure of other rules.

erfanium:
@icebob 
Can you add `Good-first-issue` tag for this?

Gamote:
@icebob I've created a PR for this #147

Gamote:
@icebob #147 PR should solve this issue, can you close it?

PR #147, opened 2020-06-05
[Feature] New validation rule: `tuple`

Gamote:
Following this discussion #124, I've created a new validation rule called: `tuple`

What was done:
- A new rule file (`tuple.js`)
- Tests (including the Typescript ones)
- README.md was updated with:
  - `tuple`'s documentation
  - `tuple`'s test coverage


Gamote:
> Great job👍. but I don't understand why tuple length is fixed to 2? It can has any length.
> for example:
> 
> ```js
> schema = {
>    $$root: true,
>    type: "tuple",
>    items: ["string", "number", "string"]
> }
> ```

@erfanium I thought about this but I had the wrong impression that a `tuple` is limited to 2 elements. I will create another commit with this changes.

Gamote:
@erfanium With the last commits a `tuple` can have any length grater than `0`. The tests and the `README.md file` were updated.

lib/rules/tuple.js
```diff
@@ -0,0 +1,67 @@
...
+	src.push(`
+		if (!Array.isArray(value)) {
+			${this.makeError({ type: "tuple", actual: "value", messages })}
+			return value;
+		}
+
+		var len = value.length;
+	`);
+
+	src.push(`
```
icebob, line 27: I don't think we should limit this new rule that it can't be empty. It's same as array, so it should be an `empty` property.

https://github.com/icebob/fastest-validator/blob/master/lib/rules/array.js#L17

erfanium:
@Gamote It's better to squash all these 28 commits!

Gamote:
@erfanium You're right, but I don't know how to this without closing this PR and create another one. Do you have any idea?

icebob:
Released in 1.5.0

Issue #116, opened 2020-03-05
Introduce optional & empty into more types


simon-fraser:
I have found a few scenarios using various types where an optional or empty flag is needed.

For example the types "url" or "email" cannot be empty or optional
Neither can the type string when using a pattern.

Gamote:
@simon-fraser Do you see any other types that needs the `empty` or `optional` property?

simon-fraser:
Great start @Gamote, I cannot say that I uncovered any other types with this issue.

Gamote:
@icebob #149, #150 and #151 PRs should solve this issue, can you close it?

PR #149, opened 2020-06-05
[Enhancement] Allow `empty = true` when using the `string` type with pattern

Gamote:
@icebob This PR will solve half of this issue: #116 

Usage (before):
```js
const schema = {
    pair: { type: "string", pattern: "^fastest", empty: true}
}

console.log(v.validate({ pair: "" }, schema)); // Fail (wrong pattern)
console.log(v.validate({ pair: "fastest-validator" }, schema)); // Valid
``` 

Usage (after):
```js
const schema = {
    pair: { type: "string", pattern: "^fastest", empty: true}
}

console.log(v.validate({ pair: "" }, schema)); // Valid
console.log(v.validate({ pair: "fastest-validator" }, schema)); // Valid
```

### DIFF
--- a/lib/rules/string.js
+++ b/lib/rules/string.js
@@ -136,9 +136,15 @@ module.exports = function checkString({ schema, messages }, path, context) {
 		if (typeof schema.pattern == "string")
 			pattern = new RegExp(schema.pattern, schema.patternFlags);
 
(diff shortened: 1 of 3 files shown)


Gamote:
@icebob Tests were added.

PR #150, opened 2020-06-09
Enhancement: empty property added to `url` rule

Gamote:
The `empty` property was added to the `url` rule (#116)

What was done:
- changes on `url.js`
- tests (including the Typescript ones)
- README.md was updated

### DIFF
--- a/lib/rules/url.js
+++ b/lib/rules/url.js
@@ -6,37 +6,37 @@ const PATTERN = /^https?:\/\/\S+/;
 
 /**	Signature: function(value, field, parent, errors, context)
  */
-module.exports = function({ schema, messages }, path, context) {
+module.exports = function ({ schema, messages }, path, context) {
 	const src = [];
 
 	src.push(`
 		if (typeof value !== "string") {
-			${this.makeError({ type: "string",  actual: "value", messages })}
+			${this.makeError({ type: "string", actual: "value", messages })}
 			return value;
 		}
 	`);
 
-	src.push(`
-		var len = value.length;
-
-		if (len === 0) {
-			if (${!schema.empty}) {
+	if (!schema.empty) {
+		src.push(`
+			if (value.length === 0) {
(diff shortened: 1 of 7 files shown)


test/rules/url.spec.js
```diff
@@ -17,7 +23,7 @@ describe("Test rule: url", () => {
 		expect(check(true)).toEqual([{ type: "string", actual: true, message }]);
 
 		message = "The '' field must be a valid URL.";
-		expect(check("")).toEqual([{ type: "url", actual: "", message }]);
+		expect(check("")).toEqual(true);
```
icebob, line 26: Good job. But my problem is that it means it's a breaking change.

erfanium: @icebob And I think It's not good to allow empty string by default

Gamote: @icebob @erfanium Hmm, I see your point of view. The main reason I've set the `schema.empty` as `true` by default was because that's the case for every rule.

Would you prefer to be `false` by default?

Gamote:
@icebob Can you review the changes?

lib/rules/url.js
```diff
@@ -20,7 +20,7 @@ module.exports = function({ schema, messages }, path, context) {
 		var len = value.length;
 
 		if (len === 0) {
```
erfanium, line 22: Better to use this (avoiding extra if-statement in src)
```js
if (!schema.empty) {
		src.push(`
		var len = value.length;

		if (len === 0) {
			return ${this.makeError({ type: "urlEmpty", actual: "value", messages })}
		}
	`);
	} else {
		src.push(`
			if (value.length === 0) return value;
		`);
	}
```

erfanium:
@Gamote: Thanks! 👍,
@icebob: LGTM

icebob:
@Gamote @erfanium thanks!

PR #156, opened 2020-06-12
Update type definitions

shawnmcknight:
This PR makes a few changes to the TS type definitions:

1. I found that a few type definitions were not being exported and thus were inaccessible.  I believe the existing format of how things were being exported was a contributing factor, so I adjusted every `type` and `interface` to `export` at the location of the definition and eliminated the exports at the end of the declaration file.  Now everything should be accessible and any new definitions added, provided they follow this pattern, will end up exported without having to add them to the follow-up exported object.
2. `ValidatorConstructorOptions` was declared as a `type`, but it would be better declared as an `interface`.  I have adjusted this type.
3. The third change I am not 100% sure on, so would love some feedback.  In the `CheckerFunctionV1` and `CheckerFunctionV2` function types, there was a `schema` parameter which was set as `ValidationSchema`.  When inspecting the parameters passed to the `custom` function, it did not appear to line up with `ValidationSchema` and instead was the rule's schema.  This seemed to represent what the code was actually doing as it passes `rule.schema` to these functions.  While I definitely think `ValidationSchema` is incorrect, I'm unsure if what should be passed in this parameter is `ValidationRule` (which is union `ValidationRuleObject | ValidationRuleObject[] | ValidationRuleName`) or a subset of that union.  In my testing, I only ever saw `ValidationRuleObject` (single, not array) being passed in, but I'll admit I'm quite unsure of what can actually get passed into the custom function.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,7 +2,7 @@ declare module "fastest-validator" {
 	/**
 	 * Type of all possible Built-in validators names
 	 */
-	type ValidationRuleName =
+	export type ValidationRuleName =
 		| "any"
 		| "array"
 		| "boolean"
@@ -28,7 +28,7 @@ declare module "fastest-validator" {
 	 * Validation schema definition for "any" built-in validator
 	 * @see https://github.com/icebob/fastest-validator#any
 	 */
-	interface RuleAny extends RuleCustom {
+	export interface RuleAny extends RuleCustom {
 		/**
 		 * Name of built-in validator
 		 */
@@ -39,7 +39,7 @@ declare module "fastest-validator" {
 	 * Validation schema definition for "array" built-in validator
 	 * @see https://github.com/icebob/fastest-validator#array
 	 */
-	interface RuleArray<T = any> extends RuleCustom {
+	export interface RuleArray<T = any> extends RuleCustom {
 		/**
 		 * Name of built-in validator
 		 */
@@ -79,7 +79,7 @@ declare module "fastest-validator" {
 	 * Validation schema definition for "boolean" built-in validator
 	 * @see https://github.com/icebob/fastest-validator#boolean
 	 */
-	interface RuleBoolean extends RuleCustom {
+	export interface RuleBoolean extends RuleCustom {
 		/**
 		 * Name of built-in validator
 		 */
@@ -95,7 +95,7 @@ declare module "fastest-validator" {
 	 * Validation schema definition for "class" built-in validator
 	 * @see https://github.com/icebob/fastest-validator#class
 	 */
-	interface RuleClass<T = any> extends RuleCustom {
+	export interface RuleClass<T = any> extends RuleCustom {
 		/**
 		 * Name of built-in validator
 		 */
@@ -110,7 +110,7 @@ declare module "fastest-validator" {
 	 * Validation schema definition for "date" built-in validator
 	 * @see https://github.com/icebob/fastest-validator#date
 	 */
-	interface RuleDate extends RuleCustom {
+	export interface RuleDate extends RuleCustom {
 		/**
 		 * Name of built-in validator
 		 */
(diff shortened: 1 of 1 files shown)


icebob:
@erfanium Could you help with it? I'm not a TS guy.

index.d.ts
```diff
@@ -831,40 +831,40 @@ declare module "fastest-validator" {
...
 		path: string,
 		parent: object | null,
 		context: Context
 	) => true | ValidationError[];
-	type CheckerFunctionV2<T = unknown> = (
+	export type CheckerFunctionV2<T = unknown> = (
 		value: T,
 		errors: CheckerFunctionError[],
-		schema: ValidationSchema,
+		ruleSchema: ValidationRule,
```
erfanium, line 851: better to be:
```ts
ruleSchema: ValidationRule | any
```
Because in can cause problem, if it's a custom rule, with custom keys

index.d.ts
```diff
@@ -831,40 +831,40 @@ declare module "fastest-validator" {
...
 		expected?: unknown;
 		actual?: unknown;
 		field?: string;
 	}
 
-	type CheckerFunctionV1<T = unknown> = (
+	export type CheckerFunctionV1<T = unknown> = (
 		value: T,
-		schema: ValidationSchema,
+		ruleSchema: ValidationRule,
```
erfanium, line 843: Same as line 851

index.d.ts
shawnmcknight: Since `ValidationRule` is a union containing `ValidationRuleObject` which is a union containing `RuleCustom` and the custom rule already has an index signature, do you think that change is needed?

Is `ValidationRule` correct for the rule schema signature?  That is, can it be any of those union types, or will it be better defined than that?

erfanium: look at this sample:
```ts
const custom: CheckerFunctionV2 = function(value , errors , schema) {
   console.log("custom -> schema.foo", schema.foo); // Error
   return value;
}

const check = v.compile({
   a: {
      type: "custom",
      foo: "bar",
      custom,
   }
});
```
I got this TypeError: `Property 'foo' does not exist on type 'ValidationRule'`

erfanium: Also you can Change `ValidationRule` to this instead: 
```ts
type ValidationRule =
	| ValidationRuleObject
	| ValidationRuleObject[]
	| ValidationRuleName
	| { [ key: string ]: any }
```

shawnmcknight: That error is because the `ValidationRule` has not been sufficiently type narrowed.  Since `schema` can be a `ValidationRuleName` or `ValidationRuleObject[]`, you can't be guaranteed access to the property.  Adding the index signature to `type ValidationRule` doesn't change that -- the type still has to be narrowed via a typeguard in order to properly access the object property.  Something like this works:
```ts
import Validator, { CheckerFunctionV2 } from 'fastest-validator';

const v = new Validator({ useNewCustomCheckerFunction: true })

const custom: CheckerFunctionV2 = function (value, errors, schema) {
  if (typeof schema === 'string') {
    console.log("custom --> schema", schema)
  } else if (Array.isArray(schema)) {
    schema.forEach((s, i) => {
      console.log(`custom --> schema[${i}]`, s.foo);
    });
  } else {
    console.log("custom -> schema.foo", schema.foo);
  }

  return value;
}

const check = v.compile({
  a: {
    type: "string",
    foo: "bar",
    custom,
  }
});
```

If the type of the `ruleSchema` can in fact be any of those types (hence why I've been asking about that), then the `custom` logic really needs to be set up to deal with the potential of getting any of the variants.

As I dig in a bit more here, I am starting to think that the type here shouldn't be `ValidationRule` but should instead of `ValidationRuleObject`.  It appears to be that the `rule.schema` passed to the custom validator always gets converted to an object and neither the string variant nor the array variant will ever get passed to the custom checker function.  If I make that change, then `foo` should be accessible on the `schema` without the need for typeguards.  Please let me know if you agree that what gets passed here will always be the object.  If so, the consumer can even do something like this, and get type safety in the custom properties:

```ts
type CustomRule = RuleCustom & { foo?: string };

const custom: CheckerFunctionV2 = function (value, errors, schema: CustomRule) {
  console.log("custom -> schema.foo", schema.foo); // string | undefined

  return value;
}
```

erfanium:
@shawnmcknight Thanks! Change are OK to me!

PR #164, opened 2020-06-16
Add plugin support

erfanium:
```js
// Plugin Side
function myPlugin(validator){
    // you can modify validator here
    // e.g.: validator.add(...)
}
// Validator Side
const v = new Validator();
v.plugin(myPlugin)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ $ npm run bench
 - [Personalised Messages](#personalised-messages)
 - [Message types](#message-types)
   - [Message fields](#message-fields)
+- [Plugins](#plugins)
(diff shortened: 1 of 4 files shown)


icebob:
What is the use-case for this feature?

erfanium:
> What is the use-case for this feature?

I have a custom rule named `objectID` that has own error messages. With plugins I can add all of them to my validator easily:
```js
v.plugin(objectID);
```  
instead of:
```js
v.add("objectID", ObjectIdFn)
Object.assign(v.messages, objectIdMessages)
// ...
```

with this pattern, also I can configure my plugins

PR #167, opened 2020-06-21
BREAKING CHANGE: update makeError method

erfanium:
With these changes, now error messages can be generated without using slow regex replace, 4x faster in `validate with wrong obj` unit! (top: new , bottom: old)
Screenshot from 2020-06-21 15-26-00

But It will breaks current `CustomCheckerFunction` error message handler. 
Any comments?

I did not delete the failed tests

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -163,30 +163,20 @@ class Validator {
 			}
 		}
 
-		const sourceCode = [
-			"var errors = [];",
-			"var field;",
-			"var parent = null;",
-		];
+		const sourceCode = [];
+		sourceCode.push(`
+			var errors = [];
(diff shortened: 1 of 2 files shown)


icebob:
I don't understand. You skipped the replaces, and now the message doesn't contain the values. It's faster but wrong

erfanium:
> I don't understand. You skipped the replaces, and now the message doesn't contain the values. It's faster but wrong

No, it's not.
MakeMessage now convert message string to a template string. e.g:
```js
"The '{field}' field must be less than or equal to {expected}."
``` 
Will be convert to:
```js
`The ${field} field must be less than or equal to ${expected}.`
```

PR #177, opened 2020-07-10
ObjectID rule

erfanium:
For validating mongodb/bson ObjectID. I'll update README.md after approvement


lib/rules/objectID.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+/**	Signature: function(value, field, parent, errors, context)
+ */
+module.exports = function({ schema, messages, index }, path, context) {
+	const src = [];
+
+	if (!context.customs[index]) context.customs[index] = { schema };
+	else context.customs[index].schema = schema;
```
erfanium, line 9: This line is not easy to test, that's why coverage decreased

erfanium:
@icebob  The main advantage of this is we can use it directly in FV schema
```js
params = {
   id: "objectID",
   name: "string",
   age: "number"
}  
```

icebob:
I mean we don't need the mongodb lib for validation. Because it's enough simple. We can copy it into the rule source.

erfanium:
@icebob But what about `convert`? It's very efficient, I use it myself

icebob:
As I see, it just convert the hex string to binary bytes with `Buffer`
https://github.com/williamkapke/bson-objectid/blob/master/objectid.js#L39-L43

icebob:
But yeah, the `instanceof` won't be correct.

icebob:
I've approved this PR. Will you update the README in this PR?

erfanium:
@icebob Yes, I do it whenever I have time, Probably the next two days, sorry for the delay

icebob:
No problem. Take your time, just I thought maybe you forgot it :)

Issue #110, opened 2020-02-04
Feature request: Currency rules


TriStarGod:
Hi,
Great library. I was wondering if you had a money / currency rule and if not, what do you use?

ishan-srivastava:
@erfanium I just stumbled upon this awesome lib.
Mind if I take this issue?

erfanium:
@ishan-srivastava Yes, You can start a pull request

ishan-srivastava:
@erfanium I have added the PR, please have a look.

PR #178, opened 2020-07-17
[feature] added rule for currency validation

ishan-srivastava:
Regarding [this](https://github.com/icebob/fastest-validator/issues/110) feature request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,6 +561,9 @@ Property | Default  | Description
 -------- | -------- | -----------
 `currencySymbol` | `null` | The currency symbol expected in string (as prefix).
 `symbolOptional` | `false` | Toggle to make the symbol optional in string, although, if present it would only allow the currencySymbol.
+`thousandSeparator` | `,` | Thousand place separator character.
+`decimalSeparator` | `.` | Decimal place character.
+`customRegex` | `null` | Custom regular expression, to validate currency strings (For eg:  /[0-9]*/g).
 
 ## `date`
 This is a `Date` validator.
--- a/README.md
+++ b/README.md
@@ -70,33 +70,34 @@ $ npm run bench
     - [Properties](#properties-1)
   - [`class`](#class)
     - [Properties](#properties-2)
-  - [`date`](#date)
+  - [`currency`](#currency)
     - [Properties](#properties-3)
-  - [`email`](#email)
+  - [`date`](#date)
     - [Properties](#properties-4)
-  - [`enum`](#enum)
+  - [`email`](#email)
     - [Properties](#properties-5)
-  - [`equal`](#equal)
+  - [`enum`](#enum)
     - [Properties](#properties-6)
-  - [`forbidden`](#forbidden)
+  - [`equal`](#equal)
     - [Properties](#properties-7)
+  - [`forbidden`](#forbidden)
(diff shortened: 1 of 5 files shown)


icebob:
@ishan-srivastava thank you for your job on this PR.
I see this rule can cover the familiar currencies like, $, €, £ but not all.
In my country, the format of currency is: `1 234,56Ft`
As you see, the thousand separator is space, the decimal separator is a comma and the sign at the end and two characters. Fully different, I know :)

image

Since you are using regex for testing, I think it would be hard to cover all types. The easier way can be that if we can set different checker regexp in the rule params. What do you think @ishan-srivastava @erfanium ?

ishan-srivastava:
Sure @icebob , that makes sense.
For what I can see is, that the validation logic can be changed by these params.
1. Currency ~~symbol~~ **string** [ having options to make it optional, and case insensitive.]
2. The currency is in prefix or postfix [default to prefix]
3. String to allow between the Currency and digits in the format of 
     a.) {**character**, minimum occurrences allowed, maximum occurrences allowed (_leave blank for infinity_) }
     b.) false if no String is allowed between the currency and digits.
4. Character for decimal [deafults to a '.'].
5. Character for thousand separator [deafults to a ',']

What we can also do is to extend the regex for the end-user to provide (in the schema), and that would override all the inbuilt validations the currency validator does, and would simply check the provided value against the regex provided by the user.

What do you think @icebob @erfanium ?

icebob:
Yeah, and please support negative currencies as well. I didn't see tests for negative values.

erfanium:
> What do you think @ishan-srivastava @erfanium ?

This is a good idea, ‌But I don't think we need to support all currencies, We can first support some important currencies and And wait for user issues/PRs 🙂

ishan-srivastava:
@erfanium sure.
I think that in order to support more currencies, we would need to add more and more logic on top of existing logic, which won't be scalable.
Like you said : Adding a param that takes in a custom currency validator name, eg currency_rial / currency_hungarian_forint, sounds good to me.
The current logic would support all the currencies that have 
a.) ',' as the thousand separator.
b.) '.' as decimal separator.
c.) Currency symbol is a single character, prefixed before the digits

Should I go ahead and add the changes I suggested, or just simply support -ve values and update the PR?

icebob:
As @erfanium said, we don't need to support all currencies. So as you described, just add some property like thousand separator and decimal separator and a custom regexp pattern. So if I want to check Hungarian forint with this rule, I will define a custom regexp and set it to the rule.

ishan-srivastava:
> Just add some property like thousand separator and decimal separator and a custom regexp pattern. 

Sure, I'll add that in this week, day job has turned v hectic :p

ishan-srivastava:
@icebob I have added the following

1. Feature to pass decimal separator
2. Feature to pass thousand separator
3. Feature to support negative currencies
4. Feature to support custom regex.

Please have a look.

erfanium:
@icebob What's the status of this issue?

icebob:
It's waiting for my review. I was busy but I will review it this week.

icebob (approved):
Good job @ishan-srivastava, thanks!

icebob:
Released in 1.7.0

PR #183, opened 2020-08-10
[BugFix] [objectID rule] validator throws error when convert: true

erfanium:


erfanium:
@icebob 
Because this bug is important but easy to fix, I'll merge these changes now without code review. Feel free to revert these changes if there's a problem

erfanium:
@icebob 
We may need to release 1.6.1 :-)

icebob:
Ok, but please wait for review in the future.

Issue #184, opened 2020-08-24
TypeError when using type: "any" 


SockeThe2nd:
I am getting `TypeError: context.fn[1] is not a function` when using type any. 

I condensed my code down to the example below, which still throws an error.
```JS
const inner = { type: "any" };

const schema = {
	prop1: inner,
	prop2: inner
};
const check = v.compile(schema);

var values = {
	prop1: "value1",
	prop2: "value2"
};

var result = check(values); 
//should be true but throws 
//TypeError: context.fn[1] is not a function
```

When using a different type instead of "any" (like "string"), the example works as expected. Additionally, when defining the schema directly (see below) the any-type works as well.

```JS
const schema = {
	prop1: { type: "any" },
	prop2: { type: "any" }
};
const check = v.compile(schema);

var values = {
	prop1: "value1",
	prop2: "value2"
};

var result = check(values); //Works as expected
```

erfanium:
@icebob Found it, I'll fix it

icebob:
Thanks, @erfanium

erfanium:
@SockeThe2nd This problem was most likely solved, Can you text again?

SockeThe2nd:
@erfanium Can confirm. My code above works with the current master-branch. Thank you very much for the quick fix.

PR #185, opened 2020-08-24
[FIX] "any" rule should have source code

erfanium:
-> #184
It is better that all the rules have source code (include `any`). Because it's difficult to handle rules that don't have source code

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -254,16 +254,11 @@ class Validator {
 
 			context.index++;
 			const res = rule.ruleFunction.call(this, rule, path, context);
-			if (res.source) {
-				res.source = res.source.replace(/%%INDEX%%/g, rule.index);
-				const fn = new Function("value", "field", "parent", "errors", "context", res.source);
-				context.fn[rule.index] = fn;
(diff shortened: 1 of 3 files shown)


lib/validator.js
```diff
@@ -254,16 +254,11 @@ class Validator {
...
 			context.index++;
 			const res = rule.ruleFunction.call(this, rule, path, context);
-			if (res.source) {
-				res.source = res.source.replace(/%%INDEX%%/g, rule.index);
-				const fn = new Function("value", "field", "parent", "errors", "context", res.source);
-				context.fn[rule.index] = fn;
-				sourceCode.push(this.wrapRequiredCheckSourceCode(rule, innerSrc.replace(/%%INDEX%%/g, rule.index), context, resVar));
-				sourceCode.push(this.makeCustomValidator({vName: resVar, path: customPath, schema: rule.schema, context, messages: rule.messages, ruleIndex: rule.index}));
-
-			} else {
```
erfanium, line 264: Now this if/else statement is useless and should be removed

icebob (approved):
Nice job. Thanks for your prompt work.

PR #190, opened 2020-09-18
array-of-x Shorthand

erfanium:


test/validator.spec.js
```diff
@@ -242,6 +242,11 @@ describe("Test getRuleFromSchema method", () => {
 			expect(res.schema).toEqual({ type: "string", empty: false, alpha: false, trim: true, some: "1234kg" });
 		});
 
+		it("should convert arrayOf syntax", () => {
+			const res = v.getRuleFromSchema("string[]");
+			expect(res.schema).toEqual({ type: "array", items: "string" });
+		});
```
icebob, line 248: And it can work like `"string[]|optional|min:1"`?

erfanium: My plan was not to support this for now, but okay, I'll test it

icebob: As I saw the code, I think it can work.

icebob (approved):
Awesome!

Issue #192, opened 2020-10-01
boolean field must be true


mtcarvalho:
I am passing one Laravel project to Moleculer.js. Now, I have to pass the Laravel [accepted](https://laravel.com/docs/5.0/validation#rule-accepted) field to Moleculer.js using fastest-validator. Is there any built-in validator that already does this? I didn't found in docs.

erfanium:
@mtcarvalho There's no such feature, but you can make your own custom rule https://github.com/icebob/fastest-validator#custom-validator

erfanium:
I personally don't like this, because JSON explicitly supports Boolean and there is no reason to use a number or string instead

icebob:
Use the `equal` rule: https://github.com/icebob/fastest-validator#equal

mtcarvalho:
thanks @icebob, this rule is exactly what I need

PR #194, opened 2020-10-07
Strict rule properties validation (WIP)

erfanium:
Related #192
I decided to use the FV itself for this!
It takes a little time because we have to create schema for each rule, For this reason, I want to check before finishing the PR whether this method is suitable or not.

### DIFF
--- a/lib/helpers/validateRuleProperties.js
+++ b/lib/helpers/validateRuleProperties.js
@@ -1,0 +1,32 @@
+let fv;
+
+const commonProps = {
+	type: "string",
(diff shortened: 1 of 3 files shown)


icebob:
Could you show before-after performance report? I hope it cannot affect the original performance.

erfanium:
@icebob 
Compile time performance or just check performance? It doesn't affect on check performance, but it slows down compilation (which is normal)

icebob:
Okay, great, yeah I know it will affect the compile time. Just I want to make sure it won't affect the check performance.

PR #196, opened 2020-10-15
allow converts objectID to string

0x0a0d:
objectID pass between service.actions should be converted to string automatically, all params of `moleculer-db-adapter-mongoose`'s actions does not accept objectID instance. If dev want an objectID instance, he should call with `convert:true`

### DIFF
--- a/lib/rules/objectID.js
+++ b/lib/rules/objectID.js
@@ -14,9 +14,11 @@ module.exports = function({ schema, messages, index }, path, context) {
 			${this.makeError({ type: "objectID", actual: "value", messages })}
 			return;
 		}
-
-		${schema.convert ? "return new ObjectID(value)" : "return value.toString();" } ;
 	`);
+
+	if (schema.convert === true) src.push("return new ObjectID(value)");
+	else if (schema.convert === "hexString") src.push("return value.toString()");
+	else src.push("return value");
 
 	return {
 		source: src.join("\n")
--- a/lib/rules/objectID.js
+++ b/lib/rules/objectID.js
@@ -15,7 +15,7 @@ module.exports = function({ schema, messages, index }, path, context) {
 			return;
 		}
 
-		${schema.convert ? "return new ObjectID(value)" : "return value;" } ;
+		${schema.convert ? "return new ObjectID(value)" : "return value.toString();" } ;
 	`);
 
 	return {
--- a/README.md
+++ b/README.md
@@ -1108,7 +1108,7 @@ const schema = {
 ### Properties
 Property | Default  | Description
 -------- | -------- | -----------
-`convert`  | `false`   | If `true`, the validator converts ObjectID HexString representation to ObjectID `instance`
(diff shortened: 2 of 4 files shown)


lib/rules/objectID.js
```diff
@@ -15,7 +15,7 @@ module.exports = function({ schema, messages, index }, path, context) {
 			return;
 		}
 
-		${schema.convert ? "return new ObjectID(value)" : "return value;" } ;
+		${schema.convert ? "return new ObjectID(value)" : "return value.toString();" } ;
```
erfanium, line 18: We can't modify the init-value when convert is false. add a property named `toString`. like:
```js
schema = {
    id: { type: "objectID", toString: true }
}
```
Also add a test

erfanium (changes requested):

lib/rules/objectID.js
0x0a0d: I though this case
let me explains:
ObjectID has 2 state: ObjectID or HexString
If we need to use ObjectID, I must sure that `convert: true` because in the case we pass a param in String or ObjectID instance, the validate code above `if (!ObjectID.isValid(value))` just make sure that param should match `/^[0-9a-f]{24}$/`. Means, if you need it exactly Instance of ObjectID or String, you must force `ObjectID(param)` or `param.toString()` before using it.
But in real, I have never met case `ObjectID(param)`, so I think it's not popular case. So convert toString should be set as default

erfanium: I personally use `ObjectID(param)`.  It's essential for those who use MongoDB native driver directly (like me).
We have also stated in the documentation that the FV doesn't modify the init-value by default. This is contrary to what the user expects.

erfanium: Also this change is breaking, We must avoid breaking changes as much as possible

0x0a0d: got it
but toString is a default object method, may I change to 
```js
schema = {
    id: { type: "objectID", string: true }
}
```

0x0a0d: an if both of `convert: true` and `string: true`, we should ignore `string` or throw an error?

erfanium: schema = {
    id: { type: "objectID", convert: "hexString" }
}

Isn't it better?

0x0a0d:
I know that but I think schema.convert will be true if you make a typo

lib/rules/objectID.js
0x0a0d: I know that but I think `schema.convert` will be true if you make a typo

icebob: I prefer @erfanium idea as using `convert: true` and `convert: "hexString"`

erfanium: @0x0a0d
```js
if (schema.convert === true) src.push("return new ObjectID(value)")
else if (schema.convert === "hexString") src.push("return value.toString()")
else src.push("return value")
```
OK?

erfanium (approved):
Thanks!

icebob (approved):

erfanium:
@icebob 
It's better to release soon, several features have been added from the previous release

PR #197, opened 2020-10-23
Update nodejs.yml

erfanium:


icebob (changes requested):
Keep the 14 because it will be the LTS, not the 15.

icebob (approved):

PR #199, opened 2020-11-08
[docs] Add supported frameworks

erfanium:
I just created `fastify-fv` plugin, It may be useful for others


icebob (approved):
Great, thanks!

PR #207, opened 2020-12-07
Remove simple usage from README.md

erfanium:
I have personally seen a large number of users using this simple method, Which is completely wrong. Also I don't think users really need a simpler method, because compiling and validating schema is crazy simple enough.


icebob (approved):
Thanks

PR #210, opened 2020-12-24
improve multi validate

0x0a0d:
in multi check process, the value passed throw validate function that can changes the original value (type 'object' with strict 'remove')
example:
```js
const Validator = require("fastest-validator");

const v = new Validator();

v.alias("targetA", {
  type: "object", strict: "remove", props: {
    a: "string"
  }
});

v.alias("targetB", {
  type: "object", strict: "remove", props: {
    b: "string"
  }
});

const schema = {
  tester: ["targetA", "targetB"]
};

v.validate({ tester: {a: "ok"} }, schema);
v.validate({ tester: {b: "not-ok"} }, schema);

```

this pull includes
+ add clone, removeChildrenAndMerge method to validator
+ add context.validator that ref to validator instance

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -419,11 +419,12 @@ class Validator {
 				`;
 			}
 
+			const result = "res_" + ruleVName;
 			return `
 				const ${ruleVName} = context.customs[${ruleIndex}];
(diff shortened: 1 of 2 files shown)


icebob (changes requested):
Please add a test which covers the original problem

PR #214, opened 2021-02-02
Base64 string validation

erfanium:
At first, I just want to support basic base64 string validations (no padding or mime options)
I used a regex from here: https://github.com/miguelmota/is-base64/blob/master/is-base64.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,9 @@ $ npm run bench
 
 # Table of contents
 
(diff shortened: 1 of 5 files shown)


icebob (approved):
Nice thanks!

erfanium:
@icebob 
May you publish it?

icebob:
Done.

PR #215, opened 2021-02-07
(docs) Update deno

### LINKED ISSUE #171: Deno + TS => Doesn't work
Because `jspm.io` doesn't support typescript, so types will not be included.
If anyone has an idea, offer it

erfanium:
fixes #171


erfanium:
@icebob 
With this new `esm.sh` CDN, FV works on deno pretty well, without types problem
I also added a basic test in a separated directory for deno to make sure FV has no problem with deno

PR #218, opened 2021-02-28
Fixed wrong arrow function syntax in documentation

ExLog:
Wrong arrow function syntax (there is no colon), changed according examples/full.js


icebob (approved):
Thx

PR #225, opened 2021-03-17
Delete `errors.push` returns as value

### LINKED ISSUE #224: Email Validation Bug - When Empty Property Is False And Value is Empty String
When schema entry for email field has "empty" property set to false, value is coerced into number.   

https://runkit.com/embed/24rdibzy73yz

```
var fastestValidator = require("fastest-validator");

const v = new fastestValidator();
const values = {someEmailField: ""};
const schema = {
    someEmailField: {
        type: "email",
        empty: false,
    }
};

const validationResult = v.validate(values, schema);
console.info({values});
```

Here is the output from runkit (link listed above):

image

erfanium:
fixes #224 
Because `Array.prototype.push` returns array length. so we shouldn't do like this:
```js
`
return ${this.makeError({...})}
`
```

instead:

```js
`
${this.makeError({...})}
return;
`
```

### DIFF
--- a/lib/rules/currency.js
+++ b/lib/rules/currency.js
@@ -1,27 +1,31 @@
 "use strict";
-const CURRENCY_REGEX = `(?=.*\\d)^(-?~1|~1-?)(([0-9]\\d{0,2}(~2\\d{3})*)|0)?(\\~3\\d{1,2})?$`;
+const CURRENCY_REGEX = "(?=.*\\d)^(-?~1|~1-?)(([0-9]\\d{0,2}(~2\\d{3})*)|0)?(\\~3\\d{1,2})?$";
 /**	Signature: function(value, field, parent, errors, context)
  */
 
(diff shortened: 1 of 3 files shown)


erfanium:
@icebob Better to release it after merge. there's some new change since last release

icebob (approved):
Thanks

erfanium:
@icebob I don't know this change should create break something or not. If someone badly used FV, may experience breaking chnage.

icebob:
Hmm could you write an example when cases can cause breaking change?

erfanium:
@icebob 
Very basic example from issue
```js
check = v.compile({
   email: { type: 'email', empty: false }
})

o = { email: "" }
check(o)

o.email // is 1 - fastest-validator@1.10.0
o.email // is "" - after this merge PR
```

Is that a breaking change? I don't think so

icebob:
No, we can merge it.

PR #226, opened 2021-03-18
support default and optional in tuples and arrays

### LINKED ISSUE #175: tuple item dosent support default and optioanl
default
```ts
const V = require('fastest-validator');

const v = new V();

const schema = {
  tuple: {
    type: 'tuple',
    empty: false,
    items: [{ type: 'string', default: '1' }],
  },
};

const data = { tuple: [] };
console.log(v.validate(data, schema));
console.log(data);
```

```text
[
  {
    type: 'tupleEmpty',
    message: "The 'tuple' field must not be an empty array.",
    field: 'tuple',
    actual: []
  }
]
{ tuple: [] }
```

optional
```ts
const schema = {
  tuple: {
    type: 'tuple',
    empty: false,
    items: [
      { type: 'string' },
      { type: 'number', optional: true },
    ],
  },
};

const data = { tuple: ['str', undefined] };
console.log(v.validate(data, schema));
console.log(data);
```

```text
[
  {
    type: 'number',
    message: "The 'tuple[1]' field must be a number.",
    field: 'tuple[1]',
    actual: undefined
  }
]
{ tuple: [ 'str', undefined ] }
```

eyalpost:
Resolves #175 
Solution: Set "value" variable to the item being validated. Similar to how properties are validated in object.js

### DIFF
--- a/lib/rules/array.js
+++ b/lib/rules/array.js
@@ -78,6 +78,7 @@ module.exports = function ({ schema, messages }, path, context) {
 			var arr = value;
 			var parentField = field;
 			for (var i = 0; i < arr.length; i++) {
+				value = arr[i];
 		`);
 
 		const itemPath = path + "[]";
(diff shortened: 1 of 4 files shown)


erfanium (approved):
Great👍

icebob (changes requested):
LGTM, but please add the these tests to the integration.spec.**js**, as well.

eyalpost:
> LGTM, but please add the these tests to the integration.spec.**js**, as well.

@icebob Done

icebob (approved):
Awesome, thanks!

PR #227, opened 2021-03-26
Update README.md

erfanium:
In this draft PR, I want to improve the docs. There are some issues that I see

- [x] `Nullable and default values` missing doc.
- [ x] ~~Explain how fastest-validator is fast and what `v.compile`  really do and why you should not call it every time.~~
- [ ] Remove `v.validate` usage from docs.


icebob:
Are you working on it, yet?

erfanium:
I'll probably finish it in the next two days

PR #228, opened 2021-03-30
Async support

icebob:
Hi,

I've added `async` support to the custom validators. It can be useful when you should check some data in the database (e.g. check unique username).

This implementation uses `async/await` because implementing a `Promise` chain would be a horrible job.
**So this new feature works only on environments which supports `async/await`**:
- Chrome > 55
- Firefox > 52
- Edge > 15
- NodeJS > 8.x (or 7.6 with harmony)

There are two new methods: `validateAsync` and `compileAsync`. These methods return `Promise` and you can use `await` keyword in the custom methods.
or you can turn on the async mode in Validation constructor options. In this case, the original `validate` and `compile` methods also return `Promise`
The two way is added to better support the usages in Moleculer, as well _(current Moleculer Validator middleware can't handle if you turn on the `async: true` in constructor)_.

**Example for 3 types of custom functions**
```js
const v = new Validator({
    debug: true,
    useNewCustomCheckerFunction: true,
    async: true
});

// Register a custom 'even' validator
v.add("even", function({ messages }) {
    return {
        source: `
            if (value % 2 != 0)
                ${this.makeError({ type: "evenNumber",  actual: "value", messages })}

            await new Promise(resolve => setTimeout(resolve, 1000));

            return value;
        `
    };
});
v.addMessage("evenNumber", "The '{field}' field must be an even number! Actual: {actual}");

const schema = {
    name: {
        type: "string",
        min: 4,
        max: 25,
        custom: async (v) => {
            await new Promise(resolve => setTimeout(resolve, 1000));
            return v.toUpperCase();
        }
    },

    username: {
        type: "custom",
        custom: async (v) => {
            // E.g. checking in the DB that whether is unique.
            await new Promise(resolve => setTimeout(resolve, 1000));
            return v.trim();
        }
    },

    age: {
        type: "even"
    }
};

(async function() {
    const data = {
        name: "John Doe",
        username: "johndoe   ",
        age: 21
    };
    console.log(await check(data), data);
})();
```

**Benchmark:**
Of course, the performance is affected in `async` mode. It should be noted in the readme.
```
Platform info:
==============
   Windows_NT 10.0.19041 x64
   Node.JS: 12.14.1
   V8: 7.7.299.13-node.16
   Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz × 8

Suite: Simple object
√ validate*                           1,220,837 rps
√ validate with sanitizations*        1,077,153 rps
√ validate with wrong obj*              679,003 rps

   validate*                              0%      (1,220,837 rps)   (avg: 819ns)
   validate with sanitizations*      -11.77%      (1,077,153 rps)   (avg: 928ns)
   validate with wrong obj*          -44.38%        (679,003 rps)   (avg: 1μs)
-----------------------------------------------------------------------
```

**Further tasks:**
- [x] Add unit tests
- [ ] Add to README


erfanium:
@icebob I had thought about this before. I agree with the generals, But there are some challenges
First, dealing with async validations without an abstraction layer is a disaster. We shouldn't use `context.async ? "await " : ""` in every single rules, they must be handled in somewhere like `validator.js`

And I think `compileAsync` is unnecessary. we return `async` checker function only when there's some async validations. (For example me as `fastify-fv` author, I can't directly call `v.compile` anymore, because I don't know should I use `compile` or `compileAsync`. this is also true in moleculer)

And `validateAsync` method also in unnecessary because we are currently deprecating `validate` method.

I am trying to work more on implementation. supporting async validation should not add more development complexity

icebob:
The reason of the `compileAsync` and `validateAsync` that I want to use the async functionality in a Moleculer services, while Moleculer core won't handle the async methods. And I don't want to change the Moleculer validator that always use the async methods because it's much slower than the sync. So with these two new methods, I can use both of sync and async functionality in the same validator instance.

Moreover, we don't need to add the `async` to the rules, because the rules contain only sync codes. Only the custom validators can be async what I covered in this PR.

erfanium:
@icebob 
This is `compileAsync` function source code
```js
compileAsync(schema) {
	return this.compile(schema, { async: true });
}
```
Why making new method when you can just call `v.compile(schema, { async: true })`? This is more like a shorthand method.

> I don't want to change the Moleculer validator that always use the async

You shouldn't use always async, moleculer middleware engine can detect async middleware, so you can do something like this: 
```js
let middleware
if (check.async) {
   middleware = async function(value) { return check(value) }
} else {
   middleware = check
}

/// using middleware
```

> because the rules contain only sync codes.

In general, user may create custom rule or custom check function. they are two different things.
for custom rules, we can ask user if you use async or not
for custom checker functions, we simply can check is provided function is `Function` or `AsyncFunction` (What you did)

icebob:
Ok, makes sense. Tomorrow I'm going to update the PR.

erfanium:
I have an idea, What do you think about taking that `async` key from `constructor` and `compile` method? We can automatically detect should schema compiles as sync or async checker function

erfanium:
That's look good, but I have a question
What is the reason for keeping `async` property in validator instance? It looks unnecessary to me

erfanium:
I will add some tests

icebob:
The idea the async in the constructor that, you can set "globally" the `async` mode in the validation instance and in this case, you don't need to add it to `validate` and `compile` methods.

erfanium:
@icebob Enabling async validating for all schemes is not a good idea and may create bad practices. I personally just create a single instance of fastest-validator for my app and async validating all schemes would cost lots of performance.

If someone needs to do it anyway, can simply extend `compile` method and override its option

icebob:
Any news?

erfanium (approved):

icebob:
Don't work on the automated detection of async?

erfanium:
@icebob 
> Don't work on the automated detection of async?

I think this is impossible. nothing works in all cases, for example returning promise in regular functions can't be detected.

> I will add some more tests

I'm a little busy these days. Don't wait for my tests, I will create a new PR. Current test are just good to be merged.

My only active opinion is about Global async option https://github.com/icebob/fastest-validator/pull/228#issuecomment-812097201. that I don't insist on it

icebob:
What if we don't try to detect it instead the developer should sign it in the schema?
**Option 1 `async: true`**
```js
    name: {
        type: "string",
        min: 4,
        max: 25,
        async: true
        custom: async (v) => {
            await new Promise(resolve => setTimeout(resolve, 1000));
            return v.toUpperCase();
        }
    },
```

**Option 2: `customAsync` method name**
```js
    name: {
        type: "string",
        min: 4,
        max: 25,
        customAsync: async (v) => {
            await new Promise(resolve => setTimeout(resolve, 1000));
            return v.toUpperCase();
        }
    },
```

What do you thing?

>My only active opinion is about Global async option #228 (comment). that I don't insist on it

I understand but I can't remove it if we don't implement a detection mode. The reason is that Moleculer users can't use the async functionality without it, because they can't add extra option to the `compile` in the Validation middleware, only they can configure the Validator global option. But if we can detect the async custom function it won't be necessary.

erfanium:
> Moleculer users can't use the async functionality without it

What about using `$$async` key in schema?
```js
schema = {
   $$async: true
   foo: "string"
}
```

icebob:
Yeah, this was the 3rd option in my head, similar to `$$root: true`.

erfanium:
> Option 2: customAsync method name
This is better, safest way with typescript

erfanium:
> Yeah, this was the 3rd option in my head, similar to $$root: true.

Also used in AJV
Yeah, this was the 3rd option in my head, similar to $$root: true.

icebob:
So I made these changes
- remove the global `async` option from Validator constructor
- remove the `opts` parameter from `validate` and `compile` methods
- add `$$async: true` property handling code

This solution much better, because you can control the async/sync mode in the schema and not in the Validator instance. It covers my Moleculer use-cases, as well.

**Example**
```js
const schema = {
    // Turn on async mode for this schema
    $$async: true,
    name: {
        type: "string",
        min: 4,
        max: 25,
        custom: async (v) => {
            await new Promise(resolve => setTimeout(resolve, 1000));
            return v.toUpperCase();
        }
    },

    username: {
        type: "custom",
        custom: async (v) => {
            // E.g. checking in the DB that whether is unique.
            await new Promise(resolve => setTimeout(resolve, 1000));
            return v.trim();
        }
    },
}
```

The compiled `check` function has an `async` property to detect this mode. If `true` it returns a `Promise`.
```js
console.log("Is async?", check.async);
```

index.d.ts
```diff
@@ -1012,7 +1018,7 @@ declare module "fastest-validator" {
 		 */
 		compile<T = any>(
```
erfanium, line 1019: Suggestion:

```ts
interface SyncCheckFunction {
   (value: any): true | ValidationError[]
   async: false
}

interface AsyncCheckFunction {
   (value: any): Promise<true | ValidationError[]>
   async: true
}

compile<T = any>(schema: ValidationSchema<T> | ValidationSchema<T>[]): SyncCheckFunction | AsyncCheckFunction
```

erfanium (approved):

index.d.ts
icebob: Thanks!

erfanium:
> This implementation uses async/await keywords. So this feature works only on environments which supports async/await:
> 
> Chrome > 55
> Firefox > 52
> Edge > 15
> NodeJS > 8.x (or 7.6 with harmony)

This is wrong, look at `validator.js`:
```js
const AsyncFunction = Object.getPrototypeOf(async function(){}).constructor;
```
This part:  `async function(){}` always executed by requiring fastest-validator. so runtime needs to support async functions, even if you don't use async validations.
In other words, fastest-validator won't work on node<8

icebob:
You are right. I fixed it in https://github.com/icebob/fastest-validator/commit/3ab24be10310487b766b2d0e84934521c5466e12
Tested with Node 6

PR #229, opened 2021-04-04
Passing meta information for custom validators

icebob:
You can pass any extra meta information for the custom validators which is available via `context.meta`.

```js
const schema = {
    name: { type: "string", custom: (value, errors, schema, name, parent, context) => {
        // Access to the meta
        return context.meta.a;
    } },
};

const res = check(obj, {
    // Passes meta information
    meta: { a: "from-meta" }
});
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,7 +8,7 @@
 			"type": "node",
 			"request": "launch",
 			"name": "Launch dev",
-			"program": "${workspaceRoot}\\examples\\custom.js"
+			"program": "${workspaceRoot}\\examples\\async.js"
 		},
 		{
 			"type": "node",
--- a/README.md
+++ b/README.md
@@ -117,8 +117,8 @@ import FastestValidator from "https://esm.sh/fastest-validator@1"
 
 const v = new FastestValidator();
 const check = v.compile({
-	name: "string",
-	age: "number",
+    name: "string",
+    age: "number",
 });
 
 console.log(check({ name: "Erf", age: 18 })); //true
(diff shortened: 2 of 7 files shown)


erfanium:
Can you give an example of a use-case?

icebob:
I need to access some runtime values which is not available at the compiling. I want to pass the Moleculer `ctx` to the custom validators to access it and get values.

erfanium:
@icebob 
You can modify validator instance:
```js
v.ctx = ctx

const schema = {
    name: { type: "string", custom: (value, errors, schema, name, parent) => {
        // this.ctx.call 
    } },
};
```
Doesn't it work?

icebob:
It does not work if there are multiple requests, because overwrite the `ctx` and the `custom` got wrong `ctx`. So the data should be passed only to the checker function. 

Btw, why do you think it can cause problem? It's a simple pointer passing. And it separated from the user-input `data`.

erfanium:
@icebob 
> Btw, why do you think it can cause problem? It's a simple pointer passing. And it separated from the user-input data.

I mean type safely. for example if you create an alias of a rule that depends on meta object, this alias may not works in all situations. There is no guarantee that the meta object always exist. And there's no type checking system to predict this problems.

icebob:
Ok, so it's a typescript issue. What if we add a generic for the context like 
```js
	export interface Context<M = unknown> {
		index: number;
		async: boolean;
		rules: ValidationRuleObject[];
		fn: Function[];
		customs: {
			[ruleName: string]: { schema: RuleCustom; messages: MessagesType };
		};
		meta?: M;
	}
```

erfanium:
@icebob Better to add a generic to validator class it self. This is not related to validator context.
```ts
const v = new Validator<Meta>()

check(value, meta) // meta argument now required
```

erfanium:
The other issue is custom checker function now has 6 arguments! This is very bad practice to have more than 2 or 3 arguments fo a function.

icebob:
I'm merging it because it doesn't affect the current logic. In a new PR, we can improve the TS definitions to better support the Context, but need help with it because I'm not a TS guy.

PR #237, opened 2021-05-23
fix debug mode

### LINKED ISSUE #236: bug in debug mode
```ts
const v = new V({ debug:true })
const schema = { x: "string" }
```
output:
```
SyntaxError: Unexpected token (2:10)
  1 | // Context.fn[0]
> 2 | function () { [native code] }

```

output without prettier:
```
// Context.fn[0]
function () { [native code] }
// Context.fn[1]
function () { [native code] }

...
```

Because `Function.prototype.bind` is native code.
Related #231

erfanium:
fixes #236


icebob (approved):
Good job, thanks.

PR #239, opened 2021-06-01
fix issue #235

### LINKED ISSUE #235: bug: TypeError: Cannot convert object to primitive value
**What?**

The string validator is in someway evaluating the objects coming as params and it breaks if it has native prototype references (e.g. toString, toJSON)

**How to reproduce it?**

```js
const Validator = require("fastest-validator");
const v = new Validator();
const schema = {    name: { type: "string" },};

console.log(check({ name: "test"})); // it works, and passes
console.log(check({ name: { "abc": 1 }})); // it works, and rejects
console.log(check({ name: { "toString": 1 }})); // it fails - TypeError: Cannot convert object to primitive value
```

**Environment:**

$ node --version
v12.16.1
macOS Catalina 10.15.16
fastest-validator version: 1.11.0

**Comments:**
I didn't inspect the code deeply, but if you have any suggestions about how to solve it, I will be glad to help!

erfanium:
fixes #235

### DIFF
--- a/lib/helpers/replace.js
+++ b/lib/helpers/replace.js
@@ -1,0 +1,8 @@
+function convertible(value) {
+	if (value === undefined) return "";
+	if (value === null) return "";
+	if (typeof value.toString === "function") return value;
+	return typeof value;
+}
+
+module.exports = (string, searchValue, newValue) => string.replace(searchValue, convertible(newValue));
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -5,8 +5,8 @@ try {
 	AsyncFunction = (new Function("return Object.getPrototypeOf(async function(){}).constructor"))();
 } catch(err) { /* async is not supported */}
 
-//const flatten = require("./helpers/flatten");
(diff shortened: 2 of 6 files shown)


lib/helpers/replace.js
```diff
@@ -0,0 +1,8 @@
+function convertible(value) {
+	if (value === undefined) return "";
+	if (value === null) return "";
+	if (typeof value.toString === "function") return value;
+	return typeof value;
```
icebob, line 5: Why return the typeof and not the value?

erfanium: Because value doesn't implement `toString` method and can't be used by `String.prototype.replace`

erfanium: for reproduce:
```js
const o1 = { a: 5 }
const o2 = { toString: 5 }
const s = "abc"
s.replace("b", o1) // worked
s.replace("b", o2) // fail

```

icebob (approved):

PR #244, opened 2021-06-07
feat: shorthand for regex

erfanium:
Sample schema using this shorthand

```js
const schema = {
   a: /^(foo|bar)$/,
   b: /[a-z]/,
   c: /^[0-9]$/,
}
```


icebob (approved):

erfanium:
@icebob 
After a long absence, it's ready

Issue #247, opened 2021-06-27
[FEATURE] Add error hadler for 'enum' validator without values


VityaSchel:
If you try to validate object with this validator:

```javascript
{ type: 'enum', foobar: ['foo', 'bar'] }
```

You'll get `TypeError: Cannot read property 'join' of undefined`

I suggest adding error handler if you are not specifying `values` key in schema

VityaSchel:
Lmao looks like maintainers doesn't give a fuck about what's going on here. I've never tried to pull-request, but maybe I'll fix this issue and open a pull request.

![image](https://user-images.githubusercontent.com/59040542/123544928-b1874380-d766-11eb-86bf-28e5ba4ff59e.png)

VityaSchel:
I hope I did everything right, if not, please redo. Like really it's just 1 line...

erfanium:
Just use typescript.
I tried to add schema language validator in https://github.com/icebob/fastest-validator/pull/194 long time ago, but it's more than doing some if-statement. Typescript has been created to solve exactly these problems.

icebob:
@VityaSchel please change your tone of voice. The maintainers work on this project free of charge and in their free time. If you are not satisfied with it, fork it, and maintain your fork by yourself.

By the way, this issue is a developer error, and not a bug in the project. Similar to this, we should check all properties in all rules to ensure that the developer doesn't make typo in their code.

VityaSchel:
> @VityaSchel please change your tone of voice. The maintainers work on this project free of charge and in their free time. If you are not satisfied with it, fork it, and maintain your fork by yourself.
> 
> By the way, this issue is a developer error, and not a bug in the project. Similar to this, we should check all properties in all rules to ensure that the developer doesn't make typo in their code.

It IS bug in the project. Error handling is a best practice, so please merge my pull request instead of closing it. I won't start a fork of this project just to fix 1 bug.

VityaSchel:
Please merge pull request or tell me how to change this 1 line I propose to prevent any bugs so you can merge it later.

VityaSchel:
> Just use typescript.
> I tried to add schema language validator in #194 long time ago, but it's more than doing some if-statement. Typescript has been created to solve exactly these problems.

[...]

PR #248, opened 2021-06-27
Error handler for no enum 'values' field

VityaSchel:
See #247


icebob:
Close regarding https://github.com/icebob/fastest-validator/issues/247#issuecomment-869185185

PR #250, opened 2021-06-29
Added error handler for no values field in enum validator + tests

VityaSchel:
#247, added tests to #248


erfanium:
@VityaSchel 
You should throw an Error in compile time, not validation time.

PR #251, opened 2021-07-08
Refactor index.d.ts and ts tests

### LINKED ISSUE #171: Deno + TS => Doesn't work

erfanium:
fixes #171

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,1063 +1,1050 @@
-declare module "fastest-validator" {
+
+export type ValidationRuleName =
+	| "any"
+	| "array"
+	| "boolean"
+	| "class"
+	| "custom"
+	| "date"
+	| "email"
+	| "enum"
+	| "equal"
+	| "forbidden"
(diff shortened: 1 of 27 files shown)


icebob (commented):
I cannot judge whether it is good or not. Can it cause a breaking issue for TS users?

erfanium:
As tests are working, it should work. Also fixes typescript problem with Deno.

icebob:
Why I get a lot of ts errors when I run `npm test` locally?
image

icebob:
It works after a `npm run build`. Why TS unit tests use the d.ts file from the dist?

erfanium:
Screenshot from 2021-07-14 22-31-53
It works for me, without `npm run build`, with node v16.2.0
If you have installed `ts-node`, or `typescript` globally, remove it

PR #269, opened 2021-09-07
fix type of contains property in string validator options

ngraef:
The typing for `contains` in the string validator options incorrectly specified a string array. The rule implementation only supports a single string.


icebob (approved):
Thanks!

Issue #240, opened 2021-06-02
Do we need rollup?


erfanium:
FV can run natively on node, so there is no need for a build process. Also bundling a browser version is not in out business, Frameworks and CDNs usually do this.

I think we do not need rollup

icebob:
It's true, but it's a breaking change, so only in 2.0

PR #275, opened 2021-10-02
Add normalize method

adam-mccormick:
This PR adds a `normalize` method to the Validator class. This method can take a schema, type, or string and resolve all shorthands and turn it into a 'normalized' form of a fastest-validator schema. This is useful for being able to inspect a schema and not have to worry about dealing with short hands, aliases, etc. 

For example: 
```
validator.normalize({ a: "string[]|optional" })
```
Will result in:
```
{
  a: {
    type: "array",
    optional: true,
    items: {
      type: "string"
    }
  }
}
```

Not a typescript guy so the ts definition of this method may not be super correct.

There's some duplication with the `getRuleFromSchema` method but I didn't want to make any changes there. Happy to revisit if required.

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -502,6 +502,72 @@ class Validator {
 		if (typeof fn !== "function") throw new Error("Plugin fn type must be function");
 		return fn(this);
 	}
+
+	/**
+	 * Normalize a schema, type or short hand definition by expanding it to a full form. The 'normalized'
+	 * form is the equivalent schema with any short hands undone. This ensure that each rule; always includes
+	 * a 'type' key, arrays always have an 'items' key, 'multi' always have a 'rules' key and objects always
+	 * have their properties defined in a 'props' key
+	 *
+	 * @param { * } value The value to normalize
+	 * @returns {Object} The normalized form of the given rule or schema
+	 */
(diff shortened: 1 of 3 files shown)


lib/validator.js
```diff
@@ -502,6 +502,72 @@ class Validator {
...
+				}
+			}
+		}
+		if(typeof value === "object") {
+			if(value.type) {
+				const config = this.normalize(value.type);
+				deepExtend(result, config, { skipIfExists: true });
+			}
+			else{
+				for (const [k, v] of Object.entries(value)) {
```
erfanium, line 563: @icebob This is an ES6 syntax, should we done #240 sooner?

icebob: It should work with rollup as well, buble should convert it to ES5.

icebob (changes requested):
Thanks for the PR. I don't like that the same logic is implemented in both places. Later we have an inconsistency state if somebody makes changes in `getRuleFromSchem` but forget to update `normalize` as well. 

Please move the duplicated parts into one method and call it from both.

adam-mccormick:
@icebob I've moved the duplication to a new method. Let me know if there are any other changes you'd like to see.

PR #284, opened 2022-02-07
check `empty: true` in string rule. Fixes #283

### LINKED ISSUE #283: Possible bug: Optional enum string fails validation on empty input
## Description

I do not know if this behaviour is by design or not but when I declare some prop like this one:

```javascript
const schema = {
      Status: {
        type: 'string',
        uppercase: true,
        enum: ['a', 'b', 'c'],
        optional: true,
        nullable: true
      }
}
validate({Status: ''}, schema)
```

the validation process fails inferring that the empty value `''` is not in the `stringEnum`, even if I've specified the `optional` and the `nullable` features. Is this by design? Or shouldn't be the null and empty string, in this case, considered as part of the enumeration or checked before validating the enum? If so it should be documented or at least using `optional` or `nullable` with `enum` should throw an exception...

icebob:
```js
const schema = {
	status: {
		type: "string",
		uppercase: true,
		enum: ["a", "b", "c"],
		empty: true
	}
};

const obj = { status: "" };

console.log(check(obj), obj);
```

### DIFF
--- a/examples/issue-283.js
+++ b/examples/issue-283.js
@@ -1,0 +1,22 @@
+const Validator = require("../index");
+
+const v = new Validator({
+	debug: true,
+	useNewCustomCheckerFunction: true,
+});
+
+const schema = {
+	status: {
+		type: "string",
+		uppercase: true,
+		enum: ["a", "b", "c"],
+		//optional: true, // checks "status === undefined"
+		//nullable: true, // checks "status === null"
+		empty: true
+	}
+};
+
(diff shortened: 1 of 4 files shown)


erfanium (approved):
I don't know why empty strings should be a valid value for enums when we can use `null` value. btw the code looks good to me

erfanium:
@icebob Also make sure `schema.empty` is false by default

PR #288, opened 2022-03-07
add context.data to d.ts

0x0a0d:
context has a reference to original data which being validated

https://github.com/icebob/fastest-validator/blob/8ec0d4f6cab692082995c688c4963ac7bd328d0d/lib/validator.js#L237


erfanium (approved):

icebob (approved):
Thanks!

PR #290, opened 2022-03-15
fix(multi): item rule has custom checker will throw error if validate…

0x0a0d:
… failed

custom checker push errors to fnCustomErrorsX that will be assigned to var errors, not _errors as expected
To avoid a breaking changes, I just keep size of errors before validate and restore it if failed
But I think the logic should be changed
image


0x0a0d:
I add some test that will break with current master branch

test/rules/multi.spec.js
```diff
@@ -88,4 +88,74 @@ describe("Test rule: multi", () => {
...
+		it("test num99", () => {
+			{
+				const o = { a: 99 };
+				expect(check(o)).toBe(true);
+				expect(o).toStrictEqual({ a: 99 });
+				expect(checkerFn).toBeCalledTimes(5);
+			}
+			{
+				const o = { a: 1199 };
+				expect(check(o)).not.toBe(true);
```
icebob, line 155: Same

icebob (changes requested):

0x0a0d:
@erfanium pls fix this too

icebob (approved):
Thanks!

PR #291, opened 2022-03-16
fix(validator): custom rules should not overwrite original value when has error

0x0a0d:


0x0a0d:
@erfanium @icebob deno CI failed because it got 404 error, not by  run tests failed. Please review and publish it if possible:)

erfanium:
@0x0a0d I fixed it

icebob:
As you see in the unit tests, it's a breaking change because it changes the expectations. 
By the way, I don't think it's an issue. The internal rules use the same logic. Sanitizations doesn't depend on errors. they can add validation errors and change the values.

icebob:
As you see in the unit tests, it's a breaking change because it changes the expectations. 
By the way, I don't think it's an issue. The internal rules use the same logic. Sanitizations doesn't depend on errors. they can add validation errors and change the values.

0x0a0d:
It's ok if we only do single test. But in multi rule test, if first test is failed, and the result is overwritten, the next test (in multi rules) will break

Issue #276, opened 2021-11-05
feature request: `map` rule


sbmelvin:
Is it possible to write a schema that validates an object whose keys aren't known, but the schema of the values are?

For example:

``` javascript
const obj = {
  steve: {
    age: 30,
    score: 100
  },
  doug: {
    age: 40,
    score: 10
  },
  bob: {
    age: 10,
    score: 3
  }
}`

In the data I'm being provided, the keys can be any valid key name but their values are objects with a consistent schema.

**Comment by icebob** [comment]
Please explain it better.

**Comment by erfanium** [comment]
@icebob 
he/she means something like this
```js
schema = {
   "*": "string"
}
```

icebob:
Ahh, thanks. For this purpose, we use custom validators.

erfanium:
@icebob I think we can introduce a new rule called `Map` for these usage.

```js
const schema = {
   dictionary: {
      type: "map",
      key: { type: "string" },
      value: { type: "object": props: { foo: "string" } }
   }
}
```

sbmelvin:
> Please explain it better.

Say I have a data source that returns the current active users of a service:
``` javascript
    const status = {
      timestamp: 1233113131,
      numUsers: 3,
      uptime: 201,
      users: {
        steve_carlos: {
          email: 'steve.carlos@ffffff.com',
          account_type: 'premium',
          last_login: 11244114
        },
        doug_flutey: {
          // same props as steve_carlos
        },
        mike_jones: {
          // same props
        }
      }
    }
```

I can't write a schema to validate this object because the keys of `users` change depending on which users are online. It's no problem to validate the objects within `users` because they have a known and unchanging schema, however validating the entire `status` object requires multiple schemas and multiple checks because there is currently no way to validate an object's props without explicitly listing the prop's name in the schema. 

So basically I need to validate an object much in the same was as an array is validated. For example, if `items` is provided instead of `props`, iterate through the keys of the object.

sbmelvin:
> @icebob I think we can introduce a new rule called `Map` for these kind of usages.
> 
> ```js
> const schema = {
>    dictionary: {
>       type: "map",
>       key: "string", // or number, default: string
>       value: { type: "object": props: { foo: "string" } }
>    }
> }
> ```

Does `type: "map"` here refer to a JS `Map`? I ask because a key in a JS `Map` can be any value (Object, function, string, primitives, etc).

erfanium:
@sbmelvin
[...]

PR #292, opened 2022-03-28
feat(rule:object): add entries to dynamic validate object key,value

0x0a0d:
add `entries` to object rule (same Object.entries) that will validate each object [key, value]
```js
{
  type: "object",
  entries: {
    key?: ValidationSchema,
    value?: ValidationSchema,
  }
}
```
===
maybe #276
need someone help me to update README because of my bad Engrisk

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -368,6 +368,13 @@ export interface RuleObject extends RuleCustom {
 	 * If set to a number, will throw if the number of props is greater than that number.
 	 */
 	maxProps?: number;
+	/**
+	 * validate object key and value
+	 */
+	entries?: {
+		key?: ValidationSchema
+		value?: ValidationSchema
+	}
 }
 
 export interface RuleObjectID extends RuleCustom {
--- a/lib/rules/object.js
+++ b/lib/rules/object.js
@@ -41,10 +41,10 @@ module.exports = function ({ schema, messages }, path, context) {
 		}
 	`);
 
+	sourceCode.push("var parentObj = value;");
+	sourceCode.push("var parentField = field;");
 	const subSchema = schema.properties || schema.props;
 	if (subSchema) {
-		sourceCode.push("var parentObj = value;");
-		sourceCode.push("var parentField = field;");
 
 		const keys = Object.keys(subSchema);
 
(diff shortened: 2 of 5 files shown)


erfanium:
why not creating new `map` rule? these changes makes object rule really complicated and hard to maintain

0x0a0d:
> why not creating new `map` rule? these changes makes object rule really complicated and hard to maintain

In my mind, `map` is applied for array only. `entries` will destructure object to [key, value] to test. If you want to `map` an array, I think we had array rule and `items` prop. Please look at my commit https://github.com/icebob/fastest-validator/pull/292/commits/a9ea63f433c08acdff6823ac67eea0af0d6bb2bf

erfanium:
@0x0a0d no `map` is not applied to array. `map` is a short name for `hashmap`. hashmap is a simple key value structure. in javascript we can have hashmaps by using literal objects `{}` or `Map` constructor.

0x0a0d:
@erfanium yes I know `Map` but `JSON.stringify` and `JSON.parse` can not transfer it. Only pure object is support

erfanium:
@0x0a0d no i mean `map` rule should handle both basic objects and `Map`s too. like current `object` rule, but the logic is implemented in a separated rule

0x0a0d:
@erfanium because it's for `object` so i put it in `object` T__T Why we must split to new rule just because it's too complicated

erfanium:
@0x0a0d you can check current `array` and `tuple` rules. both of them are for arrays but they have their own rules and implementation.

0x0a0d:
@erfanium okay you won <(")
How do you think about this
```js
{
  type: "map",
  entries: {
    key?: ValidationSchema,
    value?: ValidationSchema,
  } | Array<{
    key?: ValidationSchema,
    value?: ValidationSchema,
  }>
}
```

0x0a0d:
but I still think it's not correct for the map
example
- what if the map needs to limit minProps, maxProps?
- since maps are for objects, what if I need to make sure that some props will appear, others I need to validate by key, value?
```js
const profile = {
	name: "string",
	age: "number",
	extra_prop_1: "string",
	extra_prop_2: "string",
	extra_prop_3: "string",
	extra_prop_4: "string",
};
```

erfanium:
@icebob Can you be a decision maker here? I have no idea

icebob:
> why not creating new `map` rule? these changes makes object rule really complicated and hard to maintain

I agree with @erfanium. Thanks the PR, but the object rule would be more complex. And to check the object keys, you can use custom checkers as well.

PR #295, opened 2022-06-27
Improve Types and IntelliSense for `compile` Function

ITenthusiasm:
In the past, it was impossible to get any meaningful/helpful TypeScript IntelliSense from the `compile` function while using it. This PR narrows the types so that TS can do this properly... but it still leaves things fairly lose in case there are users who don't want stricter types at the moment.

Some spacing/semicolon changes were added by my IDE. I can remove that commit if it's undesirable.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -821,7 +821,7 @@ export type ValidationRule =
 /**
  * Definition for validation schema based on validation rules
  */
-export type ValidationSchema<T = any> = {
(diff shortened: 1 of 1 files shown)


ITenthusiasm:
I forgot to look at the tests. I'll look into this tomorrow.

icebob:
CI is broken.

icebob (changes requested):
CI is broken

Issue #297, opened 2022-06-30
Backward compatibility broken when migrating to version `1.12.0`


FFKL:
In versions prior to `1.12.0` all validation errors are returned together, but then the `strict`-related error became a priority.
You can change the `fastest-validator` version in the [example](https://codesandbox.io/s/recursing-monad-2ytp9e?file=/src/index.js) and verify it.

icebob:
The issue came with #272 by @0x0a0d here: https://github.com/icebob/fastest-validator/pull/272/files#diff-7380b3ab547c30fffa792d657dffe20479df9bcebf4932f8a97520611241cd9eR72-R74

@0x0a0d can we remove this condition?

0x0a0d:
> The issue came with #272 by @0x0a0d here: https://github.com/icebob/fastest-validator/pull/272/files#diff-7380b3ab547c30fffa792d657dffe20479df9bcebf4932f8a97520611241cd9eR72-R74
> 
> @0x0a0d can we remove this condition?

This condition cannot be completely removed as it is required for `multi rule`, unless in the `multi rule`, before call checker, the data must be `deepClone`, otherwise object `strict remove` will change it and next validation will be failed

PR #298, opened 2022-07-11
fix: issue #297

0x0a0d:
In the case object's validator found errors:
+ `strict: true` should continuously validate even
+ `strict: remove` will NOT remove the field from object


icebob (approved):
Awesome, thanks!

PR #300, opened 2022-07-24
Add Record rule

FFKL:
I added a new `record` rule. I decided to use `record` instead of `map` because it can be confused with an native `Map` object.
If everything is ok, I'll improve tests and add docs.

Related issue https://github.com/icebob/fastest-validator/issues/276

### DIFF
--- a/README.md
+++ b/README.md
@@ -928,6 +928,30 @@ check(obj); // Fail
 // ]
 ```
 
+## `record`
+This validator allows to check an object with arbitrary keys.
+
+```js
+const schema = {
+    surnameGroups: {
+        type: 'record',
+        key: { type: 'string', alpha: true },
+        value: { type: 'array', items: 'string' }
+    }
+};
+const check = v.compile(schema);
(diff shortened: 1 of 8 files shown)


erfanium (approved):
It looks good to me, btw i haven't contributed to this project for more than a year, I don't have a good mind on codes

icebob:
> Awesome, please add passed tests as well, not just negative tests.

Could you make it?

FFKL:
> > Awesome, please add passed tests as well, not just negative tests.
> 
> Could you make it?

Yes, of course. I just have a lot of work and can do only small changes. When I add tests I'll let you know ;) Today or tomorrow, I think

icebob:
Ok, take your time. Just I plan to make a new release in the next days and I would like to add it to the next version. But I will wait it :)

FFKL:
@icebob done;)

icebob (approved):
Awesome, thanks!

PR #304, opened 2022-08-24
Add: haltOnFirstError option

### LINKED ISSUE #299: Halt on Error Possible?
I don't see a way currently, but is it possible to halt the validation process as soon as a validator encounters an error? I'm using an async validator, and I'd like to be able to run basic validations (string length, format, etc), before accessing the DB (which are async custom validations in the same schema property).

If a field hits an error, I'd like the validation process to stop and return the single error instead of progressing throughout and returning additional errors as well.

avanelli:
### Description
Setting `haltOnFirstError: true` will halt checks and return immediately after the first error, if any.
Closes #299 

#### Usage
```js
const v = new Validator({haltOnFirstError: true});
```
 
### Checks

- [x] Add tests
- [x] Add examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,11 @@ console.log("Second:", check({ id: 2, name: "Adam" }));
 ```
 [Try it on Repl.it](https://repl.it/@icebob/fastest-validator-fast)
 
+If you want to halt immediately after the first error use: `haltOnFirstError`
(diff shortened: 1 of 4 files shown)


icebob (approved):
Good job, thanks! Released in 1.14

PR #314, opened 2022-11-28
convert property in array type

DenisFerrero:


lib/rules/array.js
```diff
@@ -5,6 +5,16 @@
 module.exports = function ({ schema, messages }, path, context) {
 	const src = [];
 
+	let sanitized = false;
+	if (schema.convert === true) {
```
icebob, line 9: I think we shouldn't have to wrap the `null` and `undefined` values because they are no "real" values.
The current implementation will wrap the `{ roles: null }` to `{ roles: [null] }`

test/rules/array.spec.js
```diff
@@ -131,6 +131,11 @@ describe("Test rule: array", () => {
 		expect(customFnItems).toHaveBeenNthCalledWith(1, 1, [], schema.numbers.items, "numbers[]", { numbers: [1,2] }, expect.any(Object));
 		expect(customFnItems).toHaveBeenNthCalledWith(2, 2, [], schema.numbers.items, "numbers[]", { numbers: [1,2] }, expect.any(Object));
 	});
+	
+	it("should wrap single value into array", () => {
+		const check	= v.compile({ $$root: true, type: "array", items: "string", convert: true });
+		expect(check("a")).toEqual(true);
+	});
```
icebob, line 138: Please add more test, to check the santitized value as well, and add test when the input:
- an array with value,
- an empty array
- null
- undefined

icebob (changes requested):

DenisFerrero:
Added null and undefined checks + deep tests as requested. The tests have been moved to the "Sanitizations" describe section as it was more meaningful based on the context

icebob:
Please note, I changed the nullish check to a simpler one because in this lib we aim to be the fastest, and creating an array for the check is very expensive:

image

By the way, thank you for your contribution!

icebob:
Released

Issue #296, opened 2022-06-30
Is it possible to allow `undefined` value validation but return an error for `null` value?


FFKL:
`optional` setting allows `undefined` and `null` values.
`nullable` setting allows `null` and disallows `undefined`.
What can be done for the case when you want to allow `undefined` but disallow `null`?

icebob:
Custom validator :) What is the use-case, and when need this logic?

FFKL:
@icebob This is often helpful because `null` and `undefined` is not the same :) Especially when you use typescript with strict null checking. Or if you don't want unexpected `null` in your documents in db. I think, this case should be built-in, isn't it?

erfanium:
@FFKL you can make a clean up function that removes any null key from object when you want save something to the database.

Please let's keep the validator simple

FFKL:
@erfanium No, It isn't working;) I I don't think this case is about complicating the validator. Sometimes I want `null` in my object for certain keys, sometimes I don't so I can't just use a clean up function. It's about validation logic itself. In my opinion, validator should help me in this base case.

Freezystem:
TL;DR: I also think that this behavior should be patched.

This behavior is intended and originate here: https://github.com/icebob/fastest-validator/blob/master/lib/validator.js#L119  

IMO, a field could be `optional` but specifically not `null`.  

It's **really important with mongo** for example because no or `undefined` field does not override the field value in object but `null` field value does.

As a matter of fact, in JavaScript, `null` should always be considered as a value as opposed to `undefined`.
`undefined` means not set, `null` means _empty_.  

> Considered a _bottle_ for an analogy: a _null bottle_ could be represented as an empty bottle whereas an _undefined bottle_ would just mean there is no bottle at all.  

**e.g:** consider a JS Object `o = {b: null}`, `o.a` is considered as `undefined` whereas `o.b` is considered as `null`.

So this code should validate:
```js
it("should not accept null value even if optional", () => {
  const Validator = require("fastest-validator");
  const v = new Validator();
  const schema = {val: {type: "string", nullable: false}};
  const check = v.compile(schema);

  expect(check({ val: "val" })).toBe(true);
  expect(check({ val: undefined })).toBe(true);
  expect(check({})).toBe(true);
  expect(check({ val: null })).not.toBe(true); // this returns true instead of an error
});
```

And also this one:
```js
it("should accept null value as required", () => {
    const Validator = require("fastest-validator");
    const v = new Validator();
[...]

PR #317, opened 2023-02-18
consider null as a value

Freezystem:
This PR contains breaking changes and should be reviewed with care.

All fields are considered `nullable` by default.

It allows a field to be:
 - `required` and `null`.
 - `optional` but **not** `null`.

fix https://github.com/icebob/fastest-validator/issues/296

### DIFF
--- a/test/rules/any.spec.js
+++ b/test/rules/any.spec.js
@@ -1,40 +1,75 @@
 "use strict";
 
 const Validator = require("../../lib/validator");
-const v = new Validator();
 const anyRule = require("../../lib/rules/any");
 
 describe("Test rule: any", () => {
 
-	it("should give back true anyway", () => {
-		const check = v.compile({ $$root: true, type: "any" });
-
-		expect(check(null)).toEqual(true);
-		expect(check(undefined)).toEqual([{ type: "required", actual: undefined, message: "The '' field is required." }]);
-		expect(check(0)).toEqual(true);
-		expect(check(1)).toEqual(true);
-		expect(check("")).toEqual(true);
-		expect(check("true")).toEqual(true);
-		expect(check("false")).toEqual(true);
-		expect(check([])).toEqual(true);
-		expect(check({})).toEqual(true);
-	});
-
-	it("should give back true anyway", () => {
-		const check = v.compile({ $$root: true, type: "any", optional: true });
-
-		expect(check(null)).toEqual(true);
-		expect(check(undefined)).toEqual(true);
-		expect(check(0)).toEqual(true);
-		expect(check(1)).toEqual(true);
-		expect(check("")).toEqual(true);
-		expect(check("true")).toEqual(true);
-		expect(check("false")).toEqual(true);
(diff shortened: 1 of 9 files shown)


test/rules/any.spec.js
```diff
@@ -9,7 +9,7 @@ describe("Test rule: any", () => {
 	it("should give back true anyway", () => {
 		const check = v.compile({ $$root: true, type: "any" });
 
-		expect(check(null)).toEqual([{ type: "required", actual: null, message: "The '' field is required." }]);
+		expect(check(null)).toEqual(true);
```
icebob, line 12: It's not good that this case is changed. As you wrote in the PR name `null` is a value, so the `required` rule should accept it. It should be true if `nullable: false`

icebob (changes requested):

test/rules/any.spec.js
Freezystem: Actually I'm not sure I'm following you on this point.
In my mind, the case seems good:
Here we consider that `nullable` is `true` by default as it's not explicitly set to false.
Given that `any` accept any type and with `$$root` modifier `any` accept any value, it can be anything but `undefined` which is not a value.
So when `any` type is used, `null` should be a valid value.
At the contrary, if `nullable=false` is set, `any` could be anything expect `null` or `undefined`.  
  
Or we will have to consider the default value of `nullable` to `false` but I'm scared that it will defeat the purpose and implies bigger changes to the test suite.

icebob: Yeah, you are right, sorry I was wrong, I changed the lines in my mind. By the way, I would like to avoid the breaking change....breaking FV generates breaking changes in Moleculer, it means, we can't use it only just in the next version 0.15

To cover similar issue, we add a global option which switch the old/new behaviour. Can we do something similar here?

Freezystem: No worries, I also had trouble getting my mind around it at first 😄 
I'll add to this PR a global setting to be able to switch between old and new behaviour.

Freezystem: I've also added some documentation and tests that seems relevant for specific edge cases.

4 cases are tested when `considerNullAsAValue` option is set to `true`:
 - field is `required` and `nullable`, `default` value will **NOT** apply. (default)
 - field is `required` but **NOT** `nullable`, `default` value will apply if field is `null`.
 - field is `optional` and `nullable`, `default` will **NOT** apply.
 - field is `optional` but **NOT** `nullable`, `default` value will apply if field is `null`.

Freezystem: Hi @icebob, feel free to tell me if you need me to do some other changes.

icebob: Thanks, sorry for late, I'm checking...

icebob (approved):
Good job, thanks!

Freezystem:
Hi @icebob, is it possible to publish a new release that include this feature ?

icebob:
Sorry, I forgot it. I've just released as 1.17.0

PR #320, opened 2023-04-06
added module declaration ("declare module")

andreash:
The d.ts file is missing a module declaration for typescript to work. Without the declaration, typescript does not accept, for example Validator as a constructable expression.


icebob:
Could you show an example when it causes issue?

erfanium:
and I guess this will break the module in other runtime like deno. because they don't have such a `module` concept like what node does.

PR #322, opened 2023-04-18
Add type definition for haltOnFirstError on Validator constructor | fix: #316

### LINKED ISSUE #316: haltOnFirstError is missing from the type definitions
The option `haltOnFirstError` is missing from the definition of the `Validator` constructor.

It should be added to the [ValidatorConstructorOptions](https://github.com/icebob/fastest-validator/blob/master/index.d.ts#L898).

err

hateablestream:
This PR fixes #316 

Let me know if I missed something or if we need to change anything 👌

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -913,6 +913,11 @@ export interface ValidatorConstructorOptions {
 	considerNullAsAValue?: boolean;
 
 	/**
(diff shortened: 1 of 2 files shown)


icebob (approved):
Good job, thanks!

PR #339, opened 2023-10-03
fix(types): improve typing

thib3113:
The array property `unique` was missing from the RuleArray definition

---

I've done mutiples modifications in the index.d.ts : 
 - add 2 missing `ValidationRuleName`
 - add misssing properties on array ( unique and convert )
 - removing some useless "default null" (all other values without default value doesn't have a default)
 - add default to `mode` (`quick`), and correct the default of `empty` to match doc 
 - some reorder to match the documentation
 - harmonize jsdoc

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,8 +17,10 @@ export type ValidationRuleName =
 	| "multi"
 	| "number"
 	| "object"
+	| "objectID"
 	| "record"
 	| "string"
+	| "tuple"
 	| "url"
 	| "uuid"
 	| string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -39,39 +39,43 @@ export interface RuleAny extends RuleCustom {
  * @see https://github.com/icebob/fastest-validator#array
  */
 export interface RuleArray<T = any> extends RuleCustom {
-	/**
-	 * Name of built-in validator
+    /**
+     * Name of built-in validator
+     */
+    type: "array";
+    /**
+     * If true, the validator accepts an empty array [].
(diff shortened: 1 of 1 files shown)


thib3113:
I found other problems on typing, I will try to check this . 

getRuleFromSchema => typing say it only allow `ValidationRuleName | ValidationRuleName[] | { [key: string]: unknown }`, but seems to be used with schema, and Array of schema, so maybe we can add `ValidationSchema | ValidationSchema[]`

---
typescript doesn't work when imported by ESM typescript ( typescript think you are exporting a value "default" ) .

Actually I can fix this on client side by doing : 
```typescript
import Validator, { default as ValidatorDefault } from 'fastest-validator';

type ValidatorType = ValidatorDefault.default;

const v = new (Validator as unknown as new (opts?: ValidatorConstructorOptions) => ValidatorType)();
```

but not really clean, I need to investigate more on this

thib3113:
More problems : 

Class.instanceOf [seems mandatory](https://github.com/icebob/fastest-validator/blob/master/lib/rules/class.js#L8) ( else `schema.instanceOf.name` produce an error )

icebob (approved):
Great, thanks!

thib3113:
@icebob, I will wait a little because setting ready for review, because I find another problems (like some minutes before), So if I found another typescript errors, I will add them here

thib3113:
@icebob any update on this ?

icebob (approved):

icebob:
Thanks, sorry for late.

PR #341, opened 2023-11-18
allow to add metas in the schema

### LINKED ISSUE #338: allow to add metas in the schema
Is it possible to add some metas in the schemas / rules ? 

like : 
```js
    dot: {
        $$metas: {
          openapiName: 'dot'
        },
        $$type: "object",
        x: "number", // object props here
        y: "number", // object props here
    }
```

I think it can be usefull to reuse the validation params to do others things ( like here, generating openapi from params ) . 

The keys starting with $$ can be just skipped from the validation, categorized as "system properties" ? (this can also allow to add other systems properties without breaking changes)

thib3113:
Fix #338 

I think it can be useful to add some metas in the schemas / rules
 
like : 
```js
    dot: {
        $$metas: {
          openapiName: 'dot'
        },
        $$type: "object",
        x: "number", // object props here
        y: "number", // object props here
    }
```

I think it can be usefull to reuse the validation params to do others things ( like here, generating openapi from params ) . 

The keys starting with $$ can be just skipped from the validation, categorized as "system properties" ? (this can also allow to add other systems properties without breaking changes) . 

---

I've checked with the already existing keys, to remove the metas after they are used

 - `$$async` seems to be used only in `compile` (so before `getRuleFromSchema`) / same for `$$root` / seems the same for `$$strict`
 - `$$type` seems to be used in `resolveType` (in `getRuleFromSchema`) only, and only on current schema ( it will call `getRuleFromSchema` only on sub rules ), and so I removed the $$ after 
 
 
---

I also update typescript to allow other libraries to do things like : 

```typescript
declare module 'fastest-validator' {
    interface ValidationSchemaMetaKeys {
        $$metas?: {
          foo: string;
          bar: string;
        }
    }
}
```

 
TODO : 
- [ ] tests ? (the main test is to allow all other tests, without breaking them)
- [ ] do the same on a rule

### DIFF
--- a/README.md
+++ b/README.md
@@ -1547,7 +1547,7 @@ Name                | Default text
 `dateMin`	| The '{field}' field must be greater than or equal to {expected}.
 `dateMax`	| The '{field}' field must be less than or equal to {expected}.
 `forbidden`	| The '{field}' field is forbidden.
-‍‍`email` | The '{field}' field must be a valid e-mail.
+`email` | The '{field}' field must be a valid e-mail.
 `emailEmpty` | The '{field}' field must not be empty.
 `emailMin` | The '{field}' field length must be greater than or equal to {expected} characters long.
 `emailMax` | The '{field}' field length must be less than or equal to {expected} characters long.
@@ -1570,6 +1570,20 @@ Name        | Description
 `field`     | The field name
 `expected`  | The expected value
 `actual`    | The actual value
+
+# Pass custom metas
+In some case, you will need to do something with the validation schema . 
+Like reusing the validator to pass custom settings, you can use properties starting with `$$`
+
+````typescript
+const check = v.compile({
+    $$name: 'Person',
+    $$description: 'write a description about this schema',
+    firstName: { type: "string" },
+    lastName: { type: "string" },
+    birthDate: { type: "date" }    
+});
+````
 
 # Development
 ```
--- a/index.d.ts
+++ b/index.d.ts
@@ -837,9 +837,9 @@ export type ValidationRule =
 	| ValidationRuleName;
 
 /**
- * Definition for validation schema based on validation rules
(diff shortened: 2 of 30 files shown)


thib3113:
I found some problems with this approach ... that I will need to investigate . 

First, delete will modify schema if schema is an object (reference is keep if it's an object, so, delete will delete on the reference)

Second, it seems the schema is passed to check function . It need to be passed untouched . 

So, I will need to dig more on this .

thib3113:
here is a new version, that will resolve the previous bugs ... I will continues to do tests on my side . 

@icebob did you know other automatic tests to do ? 

Also, doing tests, I remark that $$async is deleted from the source object, so, if I do : 
```javascript
const schema = {
  $$async: true,
  name: { type: "string", custom: async () => ''},
};

const check2 = v.compile(schema);

console.log(check.async); // true
console.log(check.async); // false
```

it will do differents things with the same object . Is it normal ? ( I saw on other parts of the code that you will carrefully handle the reference ... that's why it's seems strange )

thib3113:
So, @icebob . I think I'm ready . 

I've added tests on each rules checking if the changes doesn't impact the rule . 

Also, I'm testing it when testing another library (writings examples that use this modified version), and doesn't get bugs (feel free to tag me if you found a bug on this in the future)

Also, I've found that sometimes you will handle the schemas with care, and sometimes not (like deleting $$async on source object), not sure if it's normal ?

icebob:
LGTM, but could you add some sentences and example about it into the docs/README.md, please?

moreover, I should check what happened with GH actions, because CI is not executed on this PR.

thib3113:
@icebob is the little documentation added, ok/enough ?

icebob (approved):
Cool, thanks!

PR #346, opened 2024-06-19
replace deprecated method substr

FerX:
replace deprecated `substr` with `substring`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr


icebob:
Thanks!

PR #349, opened 2024-11-04
Use TSTyche to test types

### LINKED ISSUE #232: Remove all typescript tests, use tsd instead
Current typescript tests are just useless. we should just write a single test for `index.d.ts` file with 'tsd'. This is best practice for Javascript based projects with typescript support

mrazauskas:
Closes #232

The referenced issue is proposing to use `tsd` to test types. I would like to suggest to use TSTyche (https://github.com/tstyche/tstyche) instead.

It does all what `tsd` can do and much more. Here are few reasons to choose TSTyche:

- it is lightweight only 225kB (instead of 48MB);
- is able to test using specific TypeScript version (`--target 4.6,5.0,latest`, this simple);
- has `describe()` and `it()` helpers that are already used in the test files of this project;
- `--watch` mode, better performance, detailed documentation (https://tstyche.org);
- and more.

---

Here is what is done in this PR:

- added `tstyche` and removed redundant dependencies;
- included `describe()` and `it()` helpers in the test files.

With this setup TSTyche type checks the test files. Just try running:

- `npx tstyche`, to run the tests using the installed TypeScript version;
- `npx tstyche --target 4.6,5.0,latest`, to typecheck using different versions of TypeScript;
- `npx tstyche date`, to run only a single file;
- `npx tstyche date --only sanitize`, to run only a single `test()` in that file.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -27,3 +27,17 @@ jobs:
         npm run build
       env:
         CI: true
+
+  test-types:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: lts/*
+    - name: npm install, build, and test types
+      run: |
+        npm ci
+        npm run build
+        npm run test:types -- --target 4.6,5.0,latest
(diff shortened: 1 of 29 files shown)


icebob:
Self-promo library

mrazauskas:
Thanks for quick feedback.

This is a dev tool. Made for dev like you and projects like this one. It is create with focus on particular task and I believe it does great job.

It is up to you to use it or not, of course. The criteria you have made me laugh ;D Have a good day.

PR #350, opened 2025-02-08
Add support enum list into number validation

hotdream1990:
Add support enum list into number validation.
Ex:
type: { type: 'number', enum: [1, 3, 5,  7] },


icebob (changes requested):
Plz add relevan tests as well

PR #351, opened 2025-04-29
fix false negative check on UUID v7 and v8

### LINKED ISSUE #352: UUID v7 and v8 do not pass validation
Despite being perfectly valid versions, [UUIDv7](https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-7) and [UUIDv8](https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-8) are flagged as invalid by the `uuid` type validator.

The UUID validator and types needs to be updated to properly take those two versions into account.

Actual behavior:
```js
const checkUUID = v.compile({ $$root: true, type: "uuid" });
checkUUID("01965dfe-77ac-78e2-9f69-51e5ce8bd080") // not true
```

Freezystem:
As per [RFC9562](https://datatracker.ietf.org/doc/html/rfc9562#table2) uuid v7 and v8 do exists and thus are perfectly valid versions.

I'm modifying the uuid regex to be able to validate those specific versions.

fix #352

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -558,7 +558,7 @@ export interface RuleUUID extends RuleCustom {
 	/**
 	 * UUID version in range 0-6
 	 */
-	version?: 0 | 1 | 2 | 3 | 4 | 5 | 6;
+	version?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8;
 }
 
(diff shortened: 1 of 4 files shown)


icebob (approved):
Good job, thanks!

PR #354, opened 2025-10-29
Added filterUndefined params to array rule

vrag86:
I add filterUndefined params to array rule. It need when create custom items validator and skip not validated items


icebob-ai (changes requested):
Thanks for the PR, @vrag86!

The idea of filtering values from arrays is useful, but there are a few issues to address:

### Missing pieces
- **No tests** — need at least: filterUndefined removes undefined, doesn't remove null/falsy, works with items validation, disabled by default
- **No TypeScript type updates** — `RuleArray` in `index.d.ts` needs a `filterUndefined?: boolean` property
- **No README documentation** for the new option

### Design concern
A more universal approach would be better here. Instead of only filtering `undefined`, consider a more general option like `filter` that could handle multiple cases:
- Filter `undefined` values
- Filter `null` values  
- Filter both `null` and `undefined` (nullish)
- Or even a custom filter function

Something like `filter: "undefined" | "null" | "nullish"` or `filter: (value) => boolean` would be more flexible and cover more use cases.

Please consider reworking this with a broader design, adding tests, TypeScript types, and documentation.

vrag86:
I fix all

PR #362, opened 2026-04-05
Feat: Add `step` property for `number` rule

E1ecti:
Implementing `step` property maintains consistency with the HTML5 range and number input types, which utilize the step attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1671,6 +1671,7 @@ Name                | Default text
 `numberEqual`	| The '{field}' field must be equal to {expected}.
 `numberNotEqual`	| The '{field}' field can't be equal to {expected}.
 `numberInteger`	| The '{field}' field must be an integer.
+`numberStep`	| The '{field}' field must be a multiple of {expected}.
 `numberPositive`	| The '{field}' field must be a positive number.
 `numberNegative`	| The '{field}' field must be a negative number.
 `array`	| The '{field}' field must be an array.
--- a/index.d.ts
+++ b/index.d.ts
@@ -352,6 +352,10 @@ export interface RuleNumber extends RuleCustom {
 	 */
 	integer?: boolean;
 	/**
+	 * The stepping interval for the input value
+	 */
+	step?: number;
+	/**
 	 * The value must be greater than zero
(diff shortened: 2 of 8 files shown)


icebob-ai (changes requested):
Thanks for the well-structured PR — all the pieces are there (rule, messages, types, tests, README). Nice work!

However, there's a **floating point precision bug** with the modulo approach:

```js
0.3 % 0.1  // → 0.09999999999999998 (not 0!)
1.2 % 0.4  // → 0.3999999999999999
```

So `{ type: "number", step: 0.1 }` would incorrectly reject `0.3`. The current tests only use integer steps (`step: 10`), which hides this issue.

**Please fix:**
1. Handle floating point precision — e.g. using a tolerance check like `Math.abs(Math.round(value / step) - value / step) > 1e-10` or a similar approach
2. Add test cases with float steps (e.g. `step: 0.1` with values like `0.3`, `0.7`, `1.2`)
3. Consider validating that `step` is a positive number (what should `step: -5` or `step: 0` do?)

Small effort, but it'll make this feature solid. Thanks!

E1ecti:
Thanks for the feedback!

I've implemented the suggested tolerance check to handle floating-point precision issues. Regarding the `step` schema validation: while negative values don't technically break the logic, I've added a check to ensure it's a positive number.

I also updated the test suite to cover floating-point scenarios and invalid schema configurations.

E1ecti:
Previous tolerance-based approach allowed certain invalid values to pass, which was incorrect behavior for the validator.

Current implementation eliminates false positives for floating-point numbers exceeding the old `1e-10` precision limit (such as `1.00000000001`).

icebob-ai (approved):
Looks great now! The floating point handling with the integer multiplication approach is solid, schema validation catches invalid steps at compile time, and the test coverage is thorough (including float edge cases).

Minor nit: there are a few unrelated whitespace changes (double-space → single-space in `makeError` calls) — ideally those would be in a separate commit, but it's not a blocker.

Thanks for the quick turnaround and the quality work — merging! 🎉

PR #365, opened 2026-06-22
fix: ValidationSchema index signature conflict with ValidationSchemaM…

jeffrson:
…etaKeys

Assigning an object with only $$strict to ValidationSchema produces a TypeScript error under stricter module resolution (TS 6):

```
Type '{ $$strict: true }' is not assignable to type 'ValidationSchema'.
  Property '$$strict' is incompatible with index signature.
    Type 'true' is not assignable to type 'ValidationRule | undefined'.
```

The root cause: `ValidationSchemaMetaKeys` declares `$$strict?: boolean | "remove"`, but the intersection with `{ [key in keyof T]: ValidationRule | undefined }` requires every string property — including `$$strict` — to satisfy `ValidationRule | undefined`. `boolean` is not a `ValidationRule`.

The fix excludes `$$`-prefixed keys from the mapped type, since they are already covered by `ValidationSchemaMetaKeys`:

```
[key in Exclude<keyof T, `$$${string}`>]: ValidationRule | undefined;
```

This is purely a type-level change with no runtime impact.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -933,7 +933,7 @@ export type ValidationSchema<T = any> = ValidationSchemaMetaKeys & {
 	/**
 	 * List of validation rules for each defined field
 	 */
-	[key in keyof T]: ValidationRule | undefined;
+	[key in Exclude<keyof T, `$$${string}`>]: ValidationRule | undefined;
 }
 
 


jeffrson:
Closing this — unfortunately it doesn't really fix the issue :-(

The `Exclude<keyof T, \$$${string}>` approach doesn't actually help here. Since `T` defaults to `any`, the key type resolves to `string | number | symbol`, and TypeScript can't subtract a template literal pattern from the broad `string` type. `Exclude<string, `$$${string}`>` stays `string`.

The real issue is that `ValidationSchemaMetaKeys & { [key: string]: ValidationRule | undefined }` is a self-contradictory intersection — the meta properties (`$$strict: boolean`) can't satisfy the index signature. A proper fix would require widening the index signature value type to include `boolean | "remove"`, which would reduce type precision.

For now, a cast to `ValidationSchema` at the call site is the right workaround. This is actually a TypeScript design limitation, not a bug in the library.

